PR #5, opened 2023-07-17
Adjust node_id check logic

claudiulodro:
I think I've fixed this in a way that preserves the logic it was meant to have. 

Currently this line is throwing an undefined index error because if `$_GET['node_id']` is empty, then it will get the value of `$_GET['node_id']`, but of course `$_GET['node_id']` could be empty because it doesn't exist. :)

### DIFF
--- a/includes/hub/admin/class-woo.php
+++ b/includes/hub/admin/class-woo.php
@@ -158,7 +158,7 @@ abstract class Woo {
 			return null;
 		}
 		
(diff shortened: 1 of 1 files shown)


leogermani (approved):
Yes, it looks good.

node_id may:
* not exist (return null)
* be empty `...&node_id&othervar=1...` (return null)
* be `0` (valid value, continue execution)

PR #8, opened 2023-09-04
Add/donation events

miguelpeixe:


leogermani (changes requested):
We also need to update the reader in the Hub, not only in other nodes. Something similar with what we do with the reader_registered event (just do the same thing on `post_process_in_hub()` method.

I'm happy to make the changes if you want

leogermani:
@miguelpeixe I updated the PR to also process the event on the Hub. If a reader makes a donation in a node and later logs into the Hub, they should be segmented there as well.

Everything works on my test, but I'm seeing one issue and I have one question:

The issue is that not only I'm seeing the donation events, but the user is also being marked as a newsletter subscriber, even with the list ID... but it's a brand new user who only made a donation on another site...

I'm then logging into the other site, sending the OTP and verifying the user

image

What seems to be happening is that when I verify the user, I'm being subscribed to the Newsletters. I think that's an issue with the registration/login flow, but wanted to confirm with you.

The question is:

`donation_new` will also be triggered for renewals. Do we want to add every renewal to the user actions there? Do we do it for donors in a stand alone site?

leogermani:
https://github.com/Automattic/newspack-network/pull/8/commits/2962201bf287af834b2cf7909e21e7acad25c251 should fix the empty site. Not sure why I didn't do this before, but it looks safe

leogermani:
@miguelpeixe https://github.com/Automattic/newspack-network/pull/8/commits/ecb226d9225d35291bfb7ae4df882de39a621335 adds a little refactor, to make sure we always insert user the same way, and with the same meta.

I'm gonna need this in https://github.com/Automattic/newspack-network/pull/17 as well... I needed to add this to one of the PRs, added to this one because it will use it 3 times :)

miguelpeixe:
> The issue is that not only I'm seeing the donation events, but the user is also being marked as a newsletter subscriber, even with the list ID... but it's a brand new user who only made a donation on another site...

Are you under RAS, connected to ActiveCampaign, and with a configured "master list"? That's the only case where a subscription should happen (without the reader opting in).

> donation_new will also be triggered for renewals. Do we want to add every renewal to the user actions there? Do we do it for donors in a stand alone site?

Since the data event [now holds the "is_renewal" property](https://github.com/Automattic/newspack-plugin/pull/2756), we can store it as "{frequency}_renewal". WDYT?

> Do we want to register a segmentation criteria here? Are we doing this in another PR? Have you thought about it?

Didn't think of anything like that. This schema is pretty flexible and can support pretty creative criteria but I wouldn't rush it. WDYT?

leogermani:
> Are you under RAS, connected to ActiveCampaign, and with a configured "master list"? That's the only case where a subscription should happen (without the reader opting in).

No, didn't it happen to you?

> Since the data event https://github.com/Automattic/newspack-plugin/pull/2756, we can store it as "{frequency}_renewal". WDYT?

It look pretty complete, but aren't there any downsides of keeping so much data ?

> Didn't think of anything like that. This schema is pretty flexible and can support pretty creative criteria but I wouldn't rush it. WDYT?

I think people are starting to play with this plugin and some of its feature are pretty abstract. If we could come up with a very simple criteria to start with people would have a better time understanding the possibilities. For example just "is network donor"

miguelpeixe:
> > Are you under RAS, connected to ActiveCampaign, and with a configured "master list"? That's the only case where a subscription should happen (without the reader opting in).
> 
> No, didn't it happen to you?

No. My other guess would be that this localStorage entry was already there from a previous reader account you were logged in to. Switching accounts doesn't clear the storage.

leogermani:
Thank you! I'm running a few more tests on the subscription data thing

leogermani:
The newsletters subscription thing happens consistently with me all the time... on a brand anonymous window, with a brand new user... not using Active Campaign..

This is most certainly not related to this PR, but I need to understand what's going on

leogermani:
Ok, figured it out. User was being automatically subscribed to a premium newsletter on one site, and I didn't know our data library was smart enough to detect that

leogermani (approved):

PR #9, opened 2023-09-05
fix: use accepted actions on webhook registration

miguelpeixe:


leogermani:
This was done in another PR

PR #13, opened 2023-11-20
feat(ads): support network targeting key-val

miguelpeixe:
Implement support for custom targeting key-values integration with Newspack Ads.

The key-value is registered as type "pre-defined" and reportable type set to "custom dimension". The available values should be updated when saving a node or when GAM executes its setup.

### How to test

1. Check out this branch and https://github.com/Automattic/newspack-ads/pull/766
2. Set up Newspack Ads with GAM connection for the hub and their nodes
3. Visit your GAM dashboard, navigate to "Inventory -> Key Values" and confirm "site" was added and the values are set to the available nodes in the network (depending on the account used, you might also find other URLs from my own local testing)
4. Confirm the "Type" is set to "Predefined" and the "Reportable" is set to "Custom dimension"
5. Configure placements with ads and visit the front-end with GAM's debug tool by adding `?googfc` to the URL
6. Open targeting info and confirm you see the site URL as follows:
image

### DIFF
--- a/includes/class-gam.php
+++ b/includes/class-gam.php
@@ -1,0 +1,65 @@
+<?php
+/**
+ * Newspack Ads GAM Integration.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network;
+
+use Newspack_Ads\Providers\GAM\Api as GAM_API;
+use Newspack_Ads\Providers\GAM_Model;
+
+/**
+ * Integration class for Newspack Ads' GAM support.
+ */
+final class GAM {
+	/**
+	 * Initialize hooks.
+	 */
+	public static function init() {
+		add_action( 'newspack_ads_setup_gam', [ __CLASS__, 'create_targeting_keys' ] );
+		add_action( 'save_post_' . Hub\Nodes::POST_TYPE_SLUG, [ __CLASS__, 'create_targeting_keys' ] );
+		add_filter( 'newspack_ads_ad_targeting', [ __CLASS__, 'add_targeting' ], 10, 2 );
+	}
+
+	/**
+	 * Create targeting keys.
+	 */
+	public static function create_targeting_keys() {
+		if ( ! Site_Role::is_hub() ) {
+			return;
+		}
+		if ( ! class_exists( 'Newspack_Ads\Providers\GAM_Model' ) ) {
+			return;
+		}
+		$api = GAM_Model::get_api();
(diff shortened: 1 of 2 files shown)


adekbadek (approved):
Works as described!

includes/class-gam.php
```diff
@@ -0,0 +1,65 @@
...
+	 * Create targeting keys.
+	 */
+	public static function create_targeting_keys() {
+		if ( ! Site_Role::is_hub() ) {
+			return;
+		}
+		if ( ! class_exists( 'Newspack_Ads\Providers\GAM_Model' ) ) {
+			return;
+		}
+		$api = GAM_Model::get_api();
```
leogermani, line 36: Don't we need a check or try-catch here to avoid errors if the Ads plugin is not active?

includes/class-gam.php
```diff
@@ -0,0 +1,65 @@
+<?php
+/**
+ * Newspack Ads GAM Integration.
+ *
```
leogermani, line 4: Would you add a more descriptive description of the purpose of this integration please?

includes/class-initializer.php
```diff
@@ -37,9 +37,10 @@ public static function init() {
 				Rest_Authenticaton::init_node_filters();
 			}
 		}
-		
+
 		Data_Listeners::init();
 		Reader_Roles_Filter::init();
+		GAM::init();
```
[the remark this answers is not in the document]
leogermani: Also, you can move the whole class inside the Hub folder

includes/class-gam.php
miguelpeixe: This method never throws an exception. It returns false if the API is not available.

includes/class-gam.php
miguelpeixe: Added in 23e9e3674d5988920cd063b6acfe3e0d61fff76c

includes/class-initializer.php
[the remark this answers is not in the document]
miguelpeixe: Not exactly, this class should load all sites because it also implements the ad targeting, which runs on nodes and hub.

includes/class-gam.php
leogermani: oh I missed the `class_exists` check right there on my face

leogermani (approved):

PR #14, opened 2023-11-22
Feat/distributor settings

leogermani:
This PR does two things

* Changes where the Distributor customizations live

When we first implemented the "Canonical URL" option, we were only considering posts distributed from the Hub to the Nodes, so the filter was only being applied to the Nodes. With this change, this filter is also applied to the Hub.

No really need to test this, but if you want to give it a go, see the testing instructions in original pr ->  https://github.com/Automattic/newspack-network/pull/1 (add to the instructions pulling a post from the Nodes to the hub and confirming the change in the canonical url is also applied to the hub

I also fixed an issue that would prevent the selection of Hub or Default as the option for the canonical urls

* The main change of this PR -> Keep post published date

This was added as a global option, so it's on by default and there's no UI to remove the behavior.

Push a post from a site to another and confirm that the post publication date is kept from the original site (the default behavior of Distributor is to set the post publication date as the distribution date.

Do the same with a Pull.

### DIFF
--- a/includes/class-distributor-customizations.php
+++ b/includes/class-distributor-customizations.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Newspack Network Distributor Customizations.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network;
+
+/**
+ * Class to initialize the customizations we do to the Distributor plugin
+ */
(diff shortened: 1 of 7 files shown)


leogermani:
Merging to unblock some further work

PR #15, opened 2023-11-24
feat: sync author bio and meta

leogermani:
This PR adds the `author_updated` event.

When a user has one of the fields we watch for updated, it creates an event to the network.

When nodes pull the event, they will update the user information, if the user already exists. This will not create the user.

## How to test

* Setup a network with at least a hub and a node
* Create users in the sites sharing the same email
* In one of the sites, edit the user profile while watching the debug.log. Just save the profile without making any changes
* Confirm that you don't see any entries in the log referring to `Author update detected`
* Now edit the profile again and make a change to a field that we do not watch. for example nickname or billing address
* Confirm you don't see anything in the logs again
* Now make some changes to the fields we watch, like display name, website and social links
* Confirm you see them in the log
* Confirm the event is dispatched and you see them in the Event Log
* Confirm the user in the other site is updated after the event is pulled

QUESTION: should we do this only for users that can be authors of a post?

### DIFF
--- a/includes/class-author-update-watcher.php
+++ b/includes/class-author-update-watcher.php
@@ -1,0 +1,170 @@
+<?php
+/**
+ * Newspack Network watcher for updates on authors' information
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network;
+
+/**
+ * Class to handle the plugin admin pages
+ */
+class Author_Update_Watcher {
+
+	/**
+	 * Flag to indicate if the processing of the author updated event is in progress.
+	 * If true, this class won't fire any events to avoid an infinite loop in author updates.
+	 *
+	 * @var boolean
+	 */
+	public static $processing_author_updated_event = false;
+
+	/**
+	 * Holds information about the authors that were updated in this request, if any.
+	 *
+	 * @var array
+	 */
+	public static $updated_authors = [];
+
+	/**
+	 * The metadata we're watching and syncing.
+	 *
+	 * @var array
+	 */
+	public static $watched_meta = [
+		// Social Links.
+		'facebook',
+		'instagram',
+		'linkedin',
+		'myspace',
+		'pinterest',
+		'soundcloud',
+		'tumblr',
+		'twitter',
+		'youtube',
+		'wikipedia',
+
+		// Core bio.
+		'description',
+
+		// Newspack.
+		'newspack_job_title',
+		'newspack_role',
+		'newspack_employer',
+		'newspack_phone_number',
+
+		// Yoast SEO.
+		'wpseo_title',
+		'wpseo_metadesc',
+		'wpseo_noindex_author',
+		'wpseo_content_analysis_disable',
+		'wpseo_keyword_analysis_disable',
+		'wpseo_inclusive_language_analysis_disable',
+	];
+
+	/**
+	 * The user properties we're watching and syncing.
+	 *
+	 * @var array
+	 */
+	public static $user_props = [
+		'display_name',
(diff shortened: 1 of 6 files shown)


includes/incoming-events/class-author-updated.php
```diff
@@ -0,0 +1,80 @@
...
+		$data = $this->get_data();
+
+		if ( isset( $data->prop ) ) {
+			$update_array = [
+				'ID' => $existing_user->ID,
+			];
+			foreach ( $data->prop as $prop_key => $prop_value ) {
+				$update_array[ $prop_key ] = $prop_value;
+			}
+			Debugger::log( 'Updating user with data: ' . print_r( $update_array, true ) );
```
adekbadek, line 67:
```suggestion
			Debugger::log( 'Updating user with data: ' . print_r( $update_array, true ) ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_print_r

```

includes/class-author-update-watcher.php
```diff
@@ -0,0 +1,170 @@
...
+		'wpseo_inclusive_language_analysis_disable',
+	];
+
+	/**
+	 * The user properties we're watching and syncing.
+	 *
+	 * @var array
+	 */
+	public static $user_props = [
+		'display_name',
```
adekbadek, line 72: Shouldn't first and last name also be synchronised?

adekbadek (changes requested):
Works as described. To pull data, I've ran `wp cron event run newspack_network_pull_from_hub`. There's one linting issue to fix and I've left one question.

includes/class-author-update-watcher.php
leogermani: We talked about only syncing what's displayed on the front-end of the site, so display name is enough. But it's a valid question

leogermani: > To pull data, I've ran wp cron event run newspack_network_pull_from_hub

Nice, you can also do it via the UI now, there's a button there "Sync now".

And in case you've missed it, there are detailed docs about how to follow the events travelling between sites:

https://github.com/Automattic/newspack-network/blob/master/DEV_NOTES.md#troubleshooting

leogermani:
Thanks for the review.

I'll wait our discussion today about what fields to sync and also if we should do this to all roles or only a subset of roles

includes/class-author-update-watcher.php
adekbadek: > We talked about only syncing what's displayed on the front-end of the site

What prompted me to ask is that name is displayed on the user list on the admin panel. For the administrator of a hub site, it'd be nice to see the full name instead of a sad "–":

image

leogermani: ah, good call!

leogermani: DOne

leogermani:
Done. Renamed the classes to "User_update" to reflect the fact that we decided to sync meta about all users, not only authors

includes/class-user-update-watcher.php
```diff
@@ -0,0 +1,172 @@
+<?php
+/**
+ * Newspack Network watcher for updates on user' information
```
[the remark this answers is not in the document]
leogermani: WOW, this is a major issue! Not only the events are processed in the wrong order, but this can result in Nodes skipping events, because we are always querying for the most recent ones!

I committed a fix and we should get this published asap!

thanks for catching this @adekbadek !

adekbadek (approved):

PR #21, opened 2023-12-06
feat: sync user avatar

leogermani:
Syncs user avatar upon changes - using Simple Local Avatars plugin

## Testing

* Set up a network with at least two sites
* Watch the logs for changes on the `simple_local_avatar` meta: `tail -f .../logfile | grep simple_local_avatar`
* Create a new user or edit a user that does not have an avatar yet
* Set an avatar for the user
* Confirm you see `Author update detected for xxxx: meta: simple_local_avatar` on your logs
* When you save the user you will also see one entry for `simple_local_avatar_rating`
* Create a new post and assign this author to the post
* Distribute the post to another site
* Confirm the user is created on the target site, and the avatar is also created there
* On the target site, Look at the media library and spot the new avatar there as the most recent item
* Back to the first site, edit a new post and assign the same author
* Distribute this new post
* Check that the post was distributed to the target site, but note that a new attachment was not uploaded to the media library - the user still gets the same avatar.

* On either of the sites, edit the user and change the avatar
* Watch the event being added in the Event Log
* Wait for the event to be pulled in the other site
* See the avatar was synced - confirm a new image was downloaded and added to the target site media library
* Now run the same test, but deleting the avatar from the user. After the event is propagated, the avatar should be removed from the user in the target site as well

---
  - https://app.asana.com/0/0/1206041342845152

### DIFF
--- a/includes/incoming-events/class-user-updated.php
+++ b/includes/incoming-events/class-user-updated.php
@@ -9,6 +9,7 @@ namespace Newspack_Network\Incoming_Events;
 
 use Newspack_Network\User_Update_Watcher;
 use Newspack_Network\Debugger;
+use Newspack_Network\Utils\Users as User_Utils;
 
 /**
  * Class to handle the User Updated Event
@@ -76,23 +77,7 @@ class User_Updated extends Abstract_Incoming_Event {
 			}
 		}
 
-		// @TODO move this to the User Utils class when it gets merged.
-		global $simple_local_avatars;
-		if ( ! $simple_local_avatars || ! is_a( $simple_local_avatars, 'Simple_Local_Avatars' ) ) {
-			return;
-		}
-
-		$avatar_meta_key = 'simple_local_avatar';
-
-		if ( array_key_exists( $avatar_meta_key, $data->meta ) && ! empty( $data->meta[ $avatar_meta_key ]['full'] ) ) {
-			Debugger::log( 'Updating user avatar' );
-			$avatar_url = $data->meta[ $avatar_meta_key ]['full'];
-			$avatar_id  = media_sideload_image( $avatar_url, 0, null, 'id' );
-			if ( $avatar_id && is_int( $avatar_id ) ) {
-				Debugger::log( 'Avatar successfully sideloaded with ID: ' . $avatar_id );
-				$simple_local_avatars->assign_new_user_avatar( $avatar_id, $existing_user->ID );
-			}
-		}
+		User_Utils::maybe_sideload_avatar( $existing_user->ID, $data->meta );
 
 	}
 
--- a/includes/incoming-events/class-user-updated.php
+++ b/includes/incoming-events/class-user-updated.php
@@ -77,7 +77,7 @@ class User_Updated extends Abstract_Incoming_Event {
 			}
 		}
 
-		User_Utils::maybe_sideload_avatar( $existing_user->ID, $data->meta );
+		User_Utils::maybe_sideload_avatar( $existing_user->ID, $data->meta, true );
 
 	}
 
--- a/includes/incoming-events/class-user-updated.php
+++ b/includes/incoming-events/class-user-updated.php
@@ -76,6 +76,24 @@ class User_Updated extends Abstract_Incoming_Event {
 			}
 		}
 
+		// @TODO move this to the User Utils class when it gets merged.
+		global $simple_local_avatars;
+		if ( ! $simple_local_avatars || ! is_a( $simple_local_avatars, 'Simple_Local_Avatars' ) ) {
+			return;
+		}
+
+		$avatar_meta_key = 'simple_local_avatar';
+
+		if ( array_key_exists( $avatar_meta_key, $data->meta ) && ! empty( $data->meta[ $avatar_meta_key ]['full'] ) ) {
+			Debugger::log( 'Updating user avatar' );
+			$avatar_url = $data->meta[ $avatar_meta_key ]['full'];
+			$avatar_id  = media_sideload_image( $avatar_url, 0, null, 'id' );
+			if ( $avatar_id && is_int( $avatar_id ) ) {
+				Debugger::log( 'Avatar successfully sideloaded with ID: ' . $avatar_id );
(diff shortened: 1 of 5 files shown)


includes/incoming-events/class-user-updated.php
```diff
@@ -76,6 +76,24 @@ public function maybe_update_user() {
 			}
 		}
 
+		// @TODO move this to the User Utils class when it gets merged.
```
miguelpeixe, line 79: Merged in #8

miguelpeixe (changes requested):

includes/incoming-events/class-user-updated.php
leogermani: yeah... but then I realized this is only going to be used in another place after https://github.com/Automattic/newspack-network/pull/17 is merged.

Because there, we are creating the authors but we are not syncing the avatar the first time the author is distributed. So when that gets merged, then we need to make avatars also synced when the author is first created... and then we can move this

leogermani:
Lots of changes done to this PR, as merging other PRs unblocked some tasks here. Instructions updated

miguelpeixe (changes requested):
I was not able to get the distribution to retrieve the avatar.

Here's the log after distributing a new post from the users:
```
[18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/distributor-customizations/class-author-ingestion.php:81 Ingesting authors from distributed post.
[18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/distributor-customizations/class-author-ingestion.php:96 Ingesting author: author@example.com
[18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/utils/class-users.php:68 Attempting to sideload user avatar
[18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/utils/class-users.php:108 No avatar found in user data
```

This happened with an author that was already known by the other site, which was supposed to get updated with the new avatar. Also with a freshly created user with an avatar that was still unknown by the other site.

leogermani:
> I was not able to get the distribution to retrieve the avatar.
> 
> Here's the log after distributing a new post from the users:
> 
> ```
> [18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/distributor-customizations/class-author-ingestion.php:81 Ingesting authors from distributed post.
> [18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/distributor-customizations/class-author-ingestion.php:96 Ingesting author: author@example.com
> [18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/utils/class-users.php:68 Attempting to sideload user avatar
> [18-Dec-2023 20:47:27 UTC] [19663] newspack-network/includes/utils/class-users.php:108 No avatar found in user data
> ```
> 
> This happened with an author that was already known by the other site, which was supposed to get updated with the new avatar. Also with a freshly created user with an avatar that was still unknown by the other site.

Can you share what's in the `newspack_network_authors` post meta for the distributed post?

miguelpeixe:
```
array (
  0 => 
  array (
    'type' => 'wp_user',
    'id' => '7',
    'display_name' => 'Avatar Author',
    'user_email' => 'avatar@example.com',
    'user_url' => '',
    'facebook' => '',
    'instagram' => '',
    'linkedin' => '',
    'myspace' => '',
    'pinterest' => '',
    'soundcloud' => '',
    'tumblr' => '',
    'twitter' => '',
    'youtube' => '',
    'wikipedia' => '',
    'first_name' => 'Avatar',
    'last_name' => 'Author',
    'description' => '',
    'newspack_job_title' => '',
    'newspack_role' => '',
    'newspack_employer' => '',
    'newspack_phone_number' => '',
    'wpseo_title' => '',
    'wpseo_metadesc' => '',
    'wpseo_noindex_author' => '',
    'wpseo_content_analysis_disable' => '',
    'wpseo_keyword_analysis_disable' => '',
    'wpseo_inclusive_language_analysis_disable' => '',
    'simple_local_avatar_rating' => 'G',
  ),
)
```

leogermani:
@miguelpeixe  thanks. Fixed in https://github.com/Automattic/newspack-network/pull/21/commits/d8889b9868bc3cb2eada6830dfdb6747b3d8c652

Guess it's one of those "let me just adjust this little thing before pushing" ...

includes/utils/class-users.php
```diff
@@ -53,4 +55,58 @@ public static function get_or_create_user_by_email( $email, $remote_site_url, $r
...
+
+		if ( array_key_exists( $avatar_meta_key, $user_data ) && ! empty( $user_data[ $avatar_meta_key ]['full'] ) ) {
+			Debugger::log( 'Updating user avatar' );
+			$avatar_url = $user_data[ $avatar_meta_key ]['full'];
+
+			if ( ! function_exists( 'media_sideload_image' ) ) {
+				require_once ABSPATH . 'wp-admin/includes/media.php';
+				require_once ABSPATH . 'wp-admin/includes/file.php';
+				require_once ABSPATH . 'wp-admin/includes/media.php';
+				require_once ABSPATH . 'wp-admin/includes/image.php';
```
miguelpeixe, line 97: `wp-admin/includes/media.php` is added twice

miguelpeixe (changes requested):
It was still presenting the same issue so I added this to line `101` of `includes/utils/class-users.php`:
```php
			if ( \is_wp_error( $avatar_id ) ) {
				Debugger::log( 'Error sideloading avatar: ' . $avatar_id->get_error_message() );
				return false;
			}
```

And got:
```
Error sideloading avatar: A valid URL was not provided.
```

It wasn't happy with my self-signed cert. Adding a filter to bypass SSL verification did the trick.

I think adding this extra validation is important to help debug other cases in which the `media_sideload_image()` may fail.

includes/utils/class-users.php
leogermani: https://github.com/Automattic/newspack-network/pull/21/commits/00c4da889eae623339396e380fbcc7ed134ca666

leogermani:
https://github.com/Automattic/newspack-network/pull/21/commits/1723d87fd23d4d9bb1e684a01cd8d413d81a91ea

miguelpeixe (approved):

PR #25, opened 2023-12-15
Add/filter bylines

leogermani:
Filters the byline of a distributed post with the authorship information sent from the other site.

Since the authorship information is already ingested from the source site, the only thing that this filter adds are guest authors to the mix, respecting the order in which they were assigned in the original site.

### Without customizations to the theme

* Set up a network with at least two sites
* In one of the sites, create a new post and make sure it has multiple authors, mixing WP users and guest authors
* Make sure all authors have something on ther Bios and their "job titles" field
* Make sure coauthorsplus is active on both sites
* In the target site, go to Distributor > Settings, and make sure to unmark "For linked distributed posts, replace the author name and link with the original site name and link. "
* Push  (or pull) the post to the other site
* Visit WP Admin > Posts in the target site and confirm you see the post there
* Confirm that, on wp admin, you only see the WP Users figuring as post authors
* Visit the post on the front end
* Confirm you see all the authors, in the same order they appear in the original site, including guest authors
* Note: WP Users' avatars depend on this other PR to be synced -> https://github.com/Automattic/newspack-network/pull/21
* Confirm you see all the authors in the Bios section at the end of the post, and that you also see the users' job titles (except for guest authors)
* Confirm authors' links take you to their author pages in the target site, except guest authors have broken links
* Edit the post in the target site and "Unlink from original post"
* Visit the post again and check that the guest authors are gone - we are no longer filtering the byline
* Link the post again and confirm guest authors are back

### With theme customizations

* Checkout https://github.com/Automattic/newspack-theme/pull/2215
* Refresh the page
* Confirm guest authors no longer have links to their profiles
* Confirm you also see the job title for guest authors in the bio section at the end of the post

### DIFF
--- a/includes/distributor-customizations/class-author-distribution.php
+++ b/includes/distributor-customizations/class-author-distribution.php
@@ -170,6 +170,8 @@ class Author_Distribution {
 	 */
 	private static function get_guest_author_for_distribution( $guest_author ) {
 
+		global $coauthors_plus;
+
 		if ( ! is_object( $guest_author ) || ! isset( $guest_author->type ) || 'guest-author' !== $guest_author->type ) {
 			return new WP_Error( 'Error getting guest author details for distribution. Invalid Guest Author' );
 		}
@@ -188,6 +190,14 @@ class Author_Distribution {
 		// CoAuthors' guest authors have a 'website' property.
 		if ( isset( $guest_author->website ) ) {
 			$author['website'] = $guest_author->website;
+		}
+
+		// Gets the guest author avatar.
+		// We only want to send an actual uploaded avatar, we don't want to send the fallback avatar, like gravatar.
+		// If no avatar was set, let it default to the fallback set in the target site.
+		$author_avatar = $coauthors_plus->guest_authors->get_guest_author_thumbnail( $guest_author, 80 );
+		if ( $author_avatar ) {
+			$author['avatar_img_tag'] = $author_avatar;
 		}
 
 		foreach ( User_Update_Watcher::$watched_meta as $meta_key ) {
--- a/includes/distributor-customizations/class-author-distribution.php
+++ b/includes/distributor-customizations/class-author-distribution.php
@@ -140,7 +140,7 @@ class Author_Distribution {
 
 		$author = [
 			'type' => 'wp_user',
-			'id'   => $user->ID,
+			'ID'   => $user->ID,
 		];
 
 
@@ -170,27 +170,15 @@ class Author_Distribution {
 	 */
 	private static function get_guest_author_for_distribution( $guest_author ) {
 
+		// CoAuthors plugin existence was checked in get_authors_for_distribution().
 		global $coauthors_plus;
 
 		if ( ! is_object( $guest_author ) || ! isset( $guest_author->type ) || 'guest-author' !== $guest_author->type ) {
 			return new WP_Error( 'Error getting guest author details for distribution. Invalid Guest Author' );
 		}
 
-		$author = [
-			'type' => 'guest_author',
-			'id'   => $guest_author->ID,
-		];
-
-		foreach ( User_Update_Watcher::$user_props as $prop ) {
-			if ( isset( $guest_author->$prop ) ) {
-				$author[ $prop ] = $guest_author->$prop;
-			}
-		}
-
-		// CoAuthors' guest authors have a 'website' property.
-		if ( isset( $guest_author->website ) ) {
-			$author['website'] = $guest_author->website;
-		}
+		$author         = (array) $guest_author;
+		$author['type'] = 'guest_author';
 
 		// Gets the guest author avatar.
 		// We only want to send an actual uploaded avatar, we don't want to send the fallback avatar, like gravatar.
@@ -198,12 +186,6 @@ class Author_Distribution {
 		$author_avatar = $coauthors_plus->guest_authors->get_guest_author_thumbnail( $guest_author, 80 );
 		if ( $author_avatar ) {
 			$author['avatar_img_tag'] = $author_avatar;
-		}
-
-		foreach ( User_Update_Watcher::$watched_meta as $meta_key ) {
-			if ( isset( $guest_author->$meta_key ) ) {
-				$author[ $meta_key ] = $guest_author->$meta_key;
-			}
 		}
 
 		return $author;
--- a/includes/distributor-customizations/class-author-distribution.php
+++ b/includes/distributor-customizations/class-author-distribution.php
@@ -26,11 +26,36 @@ class Author_Distribution {
 	 */
 	public static function init() {
 		add_filter( 'dt_push_post_args', [ __CLASS__, 'add_author_data_to_push' ], 10, 2 );
+		add_filter( 'dt_subscription_post_args', [ __CLASS__, 'add_author_data_to_push' ], 10, 2 );
 		add_filter( 'rest_prepare_post', [ __CLASS__, 'add_author_data_to_pull' ], 10, 3 );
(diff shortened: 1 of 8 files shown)


includes/distributor-customizations/class-authorship-filters.php
```diff
@@ -0,0 +1,104 @@
...
+
+		$filtered_coauthors = [];
+
+		foreach ( $distributed_authors as $distributed_author ) {
+			if ( ! isset( $distributed_author['type'] ) ) {
+				return $coauthors;
+			}
+
+			if ( 'guest_author' === $distributed_author['type'] ) {
+				$filtered_coauthors[] = (object) $distributed_author;
```
miguelpeixe, line 71: I don't think this produces the schema expected from the `get_coauthors()` method. I'm getting a lot of warnings from functions that expect a `user_login`, `user_nicename` property and `ID` (in caps) instead of `id`:

I'm aware I'm not running https://github.com/Automattic/newspack-theme/pull/2215 in this test but I don't think we should break what is returned from the function. This can lead to a lot of unexpected things breaking.

```
PHP Warning:  Undefined property: stdClass::$user_login in /Users/miguelpeixe/devel/newspack-ads/includes/providers/gam/class-gam-model.php on line 1128
PHP Warning:  Undefined property: stdClass::$ID in newspack-theme/newspack-theme/template-parts/post/author-bio.php on line 29
PHP Warning:  Undefined property: stdClass::$user_nicename in newspack-theme/newspack-theme/template-parts/post/author-bio.php on line 29
PHP Warning:  Undefined property: stdClass::$ID in newspack-theme/newspack-theme/template-parts/post/author-bio.php on line 38
PHP Warning:  Undefined property: stdClass::$user_nicename in newspack-theme/newspack-theme/template-parts/post/author-bio.php on line 38
PHP Warning:  Undefined property: stdClass::$ID in newspack-theme/newspack-theme/template-parts/post/author-bio.php on line 39
PHP Warning:  Undefined property: stdClass::$ID in newspack-theme/newspack-theme/template-parts/post/author-bio.php on line 61
PHP Warning:  Undefined property: stdClass::$user_nicename in newspack-theme/newspack-theme/template-parts/post/author-bio.php on line 61
```

For reference, this is what this line is producing for my guest author:
```
stdClass Object
(
    [type] => guest_author
    [id] => 158
    [display_name] => John Doe
    [user_email] => john@doe.com
    [website] => https://example.com
    [first_name] => John
    [last_name] => Doe
    [description] => John is a guest author
    [newspack_job_title] =>
    [newspack_role] =>
    [newspack_employer] =>
    [newspack_phone_number] =>
)
```

miguelpeixe (changes requested):

includes/distributor-customizations/class-author-distribution.php
```diff
@@ -170,6 +170,8 @@ private static function get_wp_user_for_distribution( $user ) {
 	 */
 	private static function get_guest_author_for_distribution( $guest_author ) {
 
+		global $coauthors_plus;
```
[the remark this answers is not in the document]
leogermani: It is checked in the method that calls this method

includes/distributor-customizations/class-authorship-filters.php
leogermani: > I'm aware I'm not running https://github.com/Automattic/newspack-theme/pull/2215 in this test but I don't think we should break what is returned from the function. This can lead to a lot of unexpected things breaking.

Yes, good call, I'll replicate the object with the same properties - and some additional properties shouldn't be a problem

leogermani: Fixed in https://github.com/Automattic/newspack-network/pull/25/commits/11a3fa3eec0b2563895e2978d532e22fb3b1f469

Notice we still need to remove the contents of `user_nicename` so we know we don't want to render a link for this author. But all the properties are there.

You'll need to redistribute a post with a guest author for the changes to make effect in the new distributed post. You can use the same existent guest author though

leogermani: This is the only thing this PR doesn't resolve on its own. The theme will render a broken link for the guest author. Always.

leogermani: One idea, not very elegant, but we could:

* modify the ID of the users to something easily identifieable like `-2` or `999998989`. (or do this with the user_nicename)
* Add a filter to `author_link` and return `#` for these authors
* At least the link would not point to anywhere wrong...

WDYT?

leogermani: This is how it looks like -> https://github.com/Automattic/newspack-network/pull/30

leogermani: Moreover: I think it's a good idea to set an invalid ID for guest authors, otherwise it could easily conflict with an existing author in the site

miguelpeixe: Why not `-1`? It's a common placeholder for "not found".

leogermani: I wanted to be unique, not to get confused with anything else

miguelpeixe: What else could `-1` be?

leogermani: > What else could -1 be?
Not found :) 

It's not "not found", it's "filtered by Newspack Network"

miguelpeixe: It's "not found" on that site as an entity, but `-2` is fine.

leogermani: But what do you think about the replacement for `#`?

I think it's terrible, but at least makes this feature more self-contained in this plugin. The theme can improve the support for this feature, but it will at least work better with any theme

miguelpeixe: I think `#` is a good solution. Is linking back to the guest author page on the original site a bad idea?

leogermani: That's another option. But we discussed this on that call that day and decided against it.

We didn't want to encourage its use to get this behavior "If you want to link back to the original site use a guest author, and if you want to link to the target site, use a regular user"

miguelpeixe: Good point! `#` is solid, the theme can `$author_url !== '#'` to condition the rendering of the `<a />` and others will just have a somewhat harmless empty hash link...

leogermani: Anything else here? :)

miguelpeixe (changes requested):
The filtered array to include the distributed author is still not returning the expected values from `get_coauthors()`.

Newspack Ads, for example, relies on the `user_login` property for targeting purposes and it's missing for the distributed GA:

https://github.com/Automattic/newspack-ads/blob/cf796d6b64b91da078323c36aa9ad4ff3a9250e2/includes/providers/gam/class-gam-model.php#L1124-L1131


```
PHP Warning:  Undefined property: stdClass::$user_login in /Users/miguelpeixe/devel/newspack-ads/includes/providers/gam/class-gam-model.php on line 1128
```

I guess we could change the Newspack Ads code to use something else, but it works with regular guest authors and compatibility will still be an issue.

leogermani:
@miguelpeixe are you sure you redistributed a new post with a guest author after my changes? This was fixed in https://github.com/Automattic/newspack-network/pull/25/commits/11a3fa3eec0b2563895e2978d532e22fb3b1f469

miguelpeixe (approved):
Whoops! I tested it on an old post. Working great! Thank you for the revisions!

PR #34, opened 2024-01-15
feat: add event queue debug

leogermani:
Adds a debug section that shows the events that are scheduled to be sent from the Node to the Hub

## Testing

Perform some actions in the Node, like registering with a new user, or editing a user profile in admin.

Go to Newspack Network > Node Settings and see the new "Events queue" section. Confirm the events are there, scheduled to be sent

Captura de tela de 2024-01-15 12-42-12

Force some errors, by setting a wrong URL to the Hub, or making the Hub not accessible.

Perform a new action and wait for it to display errors

Captura de tela de 2024-01-15 12-42-35

Note 1: Not adding a "Send now" button at this moment

### DIFF
--- a/DEV_NOTES.md
+++ b/DEV_NOTES.md
@@ -30,7 +30,7 @@ For example, the `reader_registered` event will create or update a WP User in th
 
 The Hub can also be an active site in the network. So events that happen on the Hub should also be propagated as if it was just another node.
 
-But when an event is fired in the Hub, there's no need to send a webhook request to itself. So we simply listen to events fired by the Data Events API and trigger the process to persist it into the `Event_Log`. This is done by `Hub\Event_Listeners` and it basically does the same thing `Hub\Webhook` does, but it is listening to local events instead of receiving webhook requests from a Node. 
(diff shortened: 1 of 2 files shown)


miguelpeixe (approved):
The new table is working great!

Non-blocking, but I'm not sure about the use of the term "queue" here. The webhook request doesn't behave like a queue. They are independent and process whenever they can, e.g., if there's a hiccup and the webhook request is scheduled to try again in a couple of minutes, a later request might run sooner. How about "Event history"?

leogermani:
> How about "Event history"

Hmm it's not history because the have not been sent yet

* Events waiting room
* Events outbox

or simply "Recent Events"

and also change "queued" with "scheduled" in the paragraph there

leogermani:
> I like "Scheduled events" even if it still shows the most recent sent events.

https://github.com/Automattic/newspack-network/pull/34/commits/6200fbf0486cee03604b36ffee7704538e883e2c

miguelpeixe (approved):
Thanks!

PR #40, opened 2024-02-05
Feat/network pass

leogermani:
Creates the Network pass feature.

Also, this work surfaced a small issue with the plugin. The post_process_in_hub method was being called even for events that were triggered in the Hub itself. This would cause the local memberhips to get the post meta and they would become uneditable.

To be consistent, this method should not be called in such cases. Just as process_in_node is not called in node for their own events.

I slightly changed the architecture there and updated the docs. It should be good now.

** Testing

Setup

* On site A, create two membership plans
* In one of the plans, add any value to the new "Network ID" field

image

* Leave Network ID blank for the other membership plan
* Do the same on Site B, two memberships
* In one of the memberships, put the same Network ID

Now let's test the flow

* Create a new membership for a user in one of the sites - in the memberhips with a Network ID
* Confirm that a `newspack_network_woo_membership_updated` is triggered and added to the event log
* Once this event reaches the other site, confirm the user is created and a membership is granted to them in the membership plan that shares the same network id
* In the site where the membership was distributed to, confirm the link to edit the membership points to the original site, where the membership was first created
* Confirm the action links (when you hover the memberhip in the admin table) shows a message "Managed in {link}" with a link to the original membership

image

* Edit the orginal membership and change its status and confirm all status changes are properly propagated across the network
* Create a membership in the membership plan without a Network ID and confirm no events are triggered


includes/woocommerce-memberships/class-admin.php
```diff
@@ -0,0 +1,149 @@
...
+	 * @param \WP_Post $post The post object.
+	 */
+	public static function render_meta_box( $post ) {
+		$network_id = get_post_meta( $post->ID, self::NETWORK_ID_META_KEY, true );
+
+		wp_nonce_field( 'newspack_network_save_membership_plan', 'newspack_network_save_membership_plan_nonce' );
+		?>
+		<label for="newspack-network-id"><?php esc_html_e( 'Network ID', 'newspack-network' ); ?></label>
+		<input type="text" id="newspack-network-id" name="newspack_network_id" value="<?php echo esc_attr( $network_id ); ?>" />
+		<p><?php esc_html_e( 'If a Network ID is informed, the user memberships will be propagated to the network. The user will be granted with the membership with the matching Network ID in all other sites in the network.', 'newspack-network' ); ?></p>
```
adekbadek:
```suggestion
		<p><?php esc_html_e( 'If a Network ID is set, the user memberships will be propagated on the network. Users will be granted the membership with the matching Network ID in all other sites in the network.', 'newspack-network' ); ?></p>
```

includes/woocommerce-memberships/class-admin.php
```diff
@@ -0,0 +1,149 @@
...
+	 *                          'Delete Permanently', 'Preview', and 'View'.
+	 * @param WP_Post  $post    The post object.
+	 * @return array
+	 */
+	public static function post_row_actions( $actions, $post ) {
+		$managed_by_newspack_network = get_post_meta( $post->ID, self::NETWORK_MANAGED_META_KEY, true );
+		if ( ! $managed_by_newspack_network ) {
+			return $actions;
+		}
+		$link     = get_edit_post_link( $post->ID );
```
adekbadek: The ID here is on the "receiving" site, not the site where the membership is managed, so the link will be wrong.

includes/woocommerce-memberships/class-admin.php
```diff
@@ -0,0 +1,149 @@
...
+	 * Filters the row actions for user memberships managed in another site of the network.
+	 *
+	 * @param string[] $actions An array of row action links. Defaults are
+	 *                          'Edit', 'Quick Edit', 'Restore', 'Trash',
+	 *                          'Delete Permanently', 'Preview', and 'View'.
+	 * @param WP_Post  $post    The post object.
+	 * @return array
+	 */
+	public static function post_row_actions( $actions, $post ) {
+		$managed_by_newspack_network = get_post_meta( $post->ID, self::NETWORK_MANAGED_META_KEY, true );
```
adekbadek, line 152: This will also be true on the membership-originating site, resulting in a confusing "Managed in <this-site>" message.

includes/woocommerce-memberships/class-admin.php
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	/**
+	 * Filters the edit post link for user memberships managed in another site of the network.
+	 *
+	 * @param string $link The edit post link.
+	 * @param int    $post_id The post ID.
+	 * @return string
+	 */
+	public static function get_edit_post_link( $link, $post_id ) {
```
adekbadek, line 132: The membership on the "receiving" site is still editable if visited directly (manually typing the post edit link). The post-save redirect goes to the wrong link, though (see comment on `post_row_actions` method). The membership updated this way (from the receiving site) is not updated on the origin site. This might be confusing. 

If the update is supposed to be one-way, the membership edit view should not be editable, just like post edit view of a distributed post.

adekbadek (changes requested):

leogermani:
Thanks @adekbadek ,

I have fixed the link and changed the approach to block edition. I have now a check for the `edit_post` capability check. See what you think.

Also, this work surfaced a small issue with the plugin. The `post_process_in_hub` method was being called even for events that were triggered in the Hub itself. This would cause the local memberhips to get the post meta and they would become uneditable.

To be consistent, this method should not be called in such cases. Just as `process_in_node` is not called in node for their own events.

I slightly changed the architecture there and updated the docs. It should be good now.

(btw, I'm open to revisiting the method names)

includes/woocommerce-memberships/class-admin.php
adekbadek: @leogermani WDYT about these wording changes?

includes/woocommerce-memberships/class-admin.php
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Handles admin tweaks for woocommerce memberships.
+ *
+ * Adds a metabox to the membership plan edit screen to allow the user to add a network id metadata to the plans
+ */
+class Admin {
+
+	/**
+	 * The membership plan custom post type.
```
adekbadek:
```suggestion
	 * The membership plan custom post type (defined in the woocommerce-memberships plugin).
```

adekbadek (approved):
Pre-approving, pending some wording suggestions.

leogermani:
@adekbadek this is good to land. But leaving this un-merged to see how we want to proceed with the release, If we want to have this in an epic branch, a feature flag... or if it's fine since it will have no effect until a Network ID is set

adekbadek:
I think we're good to go, just remember to add public-facing docs after the release.

PR #41, opened 2024-02-06
Fix Yoast primary category

adekbadek:
### All Submissions:

With this change, Yoast's primary category will be interpreted on the Node site. Even though the taxonomies are sync'd between sites by Distributor, Yoast's primary category – stored as post meta – was not handled, and its ID still pointed to the ID of the category on the Hub site. 

See 1200550061930446-as-1206399433499334/f

1. On the hub site, create a new category and publish a post after setting the new category as the primary category
2. Observe that the distributed post on the Node site has the correct primary category set*
3. Create yet another new category and update that same post so now it has this new category as the primary category 
4. Again observe that the distribute post on the Node site has the correct primary category set

\* this is stored as the `_yoast_wpseo_primary_category` post meta

### Other information:

### DIFF
--- a/includes/distributor-customizations/global.php
+++ b/includes/distributor-customizations/global.php
@@ -29,13 +29,9 @@ function newspack_network_get_primary_category_slug( $post_body, $post ) {
  * @param WP_REST_Request $request The request data.
  */
 function newspack_network_fix_primary_category( $post, $request ) {
-	$primary_category_id   = get_post_meta( $post->ID, '_yoast_wpseo_primary_category', true );
 	$primary_category_slug = get_post_meta( $post->ID, 'yoast_primary_category_slug', true );
-	$hub_primary_category  = get_term( $primary_category_id );
-	if ( ! $hub_primary_category ) {
-		// Attempt to find a matching category on the Hub site by slug.
-		$hub_primary_category = get_term_by( 'slug', $primary_category_slug, 'category' );
-	}
+	// Match the category by slug, the IDs might have a clash.
+	$hub_primary_category = get_term_by( 'slug', $primary_category_slug, 'category' );
 	if ( $hub_primary_category ) {
 		update_post_meta( $post->ID, '_yoast_wpseo_primary_category', $hub_primary_category->term_id );
 	} elseif ( class_exists( '\Newspack\Logger' ) ) {
--- a/includes/distributor-customizations/global.php
+++ b/includes/distributor-customizations/global.php
@@ -13,7 +13,7 @@
  */
 function newspack_network_get_primary_category_slug( $post, $is_pulling = false ) {
 	if ( $is_pulling ) {
-		// When pulling content, the post will be the Node post.
+		// When pulling content, the post will be the remote site post (not on the WP instance that executes this code).
 		// The category slug has to be read from the data on the post object.
 		if ( ! isset( $post->meta['_yoast_wpseo_primary_category'] ) ) {
 			return;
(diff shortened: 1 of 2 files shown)


leogermani:
Didn't work for me.

I've tested both pushing and pulling posts and the value of the `_yoast_wpseo_primary_category` post meta still shows the cat ID as it is in the site where the post was originally published.

Unrelated to this PR, once we get this merged I'll get this file refactor.

To have a file like this seemed like a good idea for small snippets, to avoid all the overhead of new files and classes for simple and small tweaks. But it's starting to grow too much. Let's break this into a class for each thing...

adekbadek:
> once we get this merged I'll get this file refactor.

Already done in https://github.com/Automattic/newspack-network/pull/42

> Didn't work for me.

I'll work on JN test site(s), so it can be debugged.

adekbadek:
@leogermani – should be fixed with 1da862c0f84ac8401113f10699bcf47a709c0f58

leogermani:
@adekbadek It fixed things for when pushing a post, but not for pulling. You also need to add the cat slug to `dt_pull_post_args` filter.

> once we get this merged I'll get this file refactor.
> Already done in https://github.com/Automattic/newspack-network/pull/42

Nice! I'll go even further and will create one file/class for each thing

adekbadek:
Thanks @leogermani ! 94030533b66658265bb57ee2c63a59d86e3dd9d4 should fix it 👌

leogermani:
Sorry, but it didn't work yet. Pulling content still gives me the wrong (original) cat id

includes/distributor-customizations/global.php
```diff
@@ -1,23 +1,82 @@
...
- * Keep the publication date on the new pushed post.
+ * Send primary category slug to the Node along with post update data.
  *
- * This filter is used to filter the arguments sent to the remote server during a push. The below code snippet passes the original published date to the new pushed post and sets the same published date instead of setting it as per the current time.
+ * @param WP_Post $post The post object.
+ * @param bool    $is_pulling Whether the post is being pulled from the remote site.
+ */
+function newspack_network_get_primary_category_slug( $post, $is_pulling = false ) {
+	if ( $is_pulling ) {
+		// When pulling content, the post will be the Node post.
```
leogermani: The "Hub" and "Node" terms don't apply here. Posts can be distributed from any site to any site (node to node, node to hub, etc).

It's better to say the origin and target sites, or something like that

adekbadek:
@leogermani – 3786e464abe370ea323f6ba7a943b238eeb82d68 should do it

includes/distributor-customizations/global.php
```diff
@@ -1,23 +1,81 @@
...
+		}
+	}
+}
+
+/**
+ * Fix primary category on the Node.
+ *
+ * @param WP_Post $post The post object.
+ */
+function newspack_network_fix_primary_category( $post ) {
```
leogermani, line 52: nit: remove the "hub" jargon from this method

includes/distributor-customizations/global.php
```diff
@@ -1,23 +1,81 @@
...
+	$primary_category_slug = get_post_meta( $post->ID, 'yoast_primary_category_slug', true );
```
leogermani: nit: I'd add a newspack_network prefix to this meta name (and maybe make it shorter. it confuses with the actual yoast meta.

leogermani (approved):
It works. 

I tested all possible combinations, and also confirmed that the category is created in the site before we attempt to fix the metadata if the cat doesnt exist.

I left two very minor comments.

includes/distributor-customizations/global.php
adekbadek: Done in 29ee85fa914a5c0e8688d28787fbed8dbebdee0d

includes/distributor-customizations/global.php
adekbadek: Done in 29ee85fa914a5c0e8688d28787fbed8dbebdee0d

PR #42, opened 2024-02-06
Sync post status

adekbadek:
### All Submissions:

Adds post status synchronization. The functional change itself is small (ffec10916321afa191dcdb1d5ac859a9ad57fa19), most of the diff is a refactor converting the global/base Distributor hooks file to a class (af03b03f5a946ecb1e8a2bb0f11fa5d748d6b56c). 

See 1205909204837811-as-1206399433499327/f

1. In the Hub site, publish a post and distribute it
2. Observe it's been distributed as expected on the Node site
3. Trash the post on the Hub 
4. Observe it's trashed on the Node
5. For a good measure, feel free to test other statuses

### Other information:


leogermani:
@adekbadek The changes look good and work fine, but I want to ask you opinion on whether we should this syncing in such a broad way.

The feature request mentioned specifically the use case of trashing a post.

I'm afraid that if we do it in this way we might introduce some unintentional behaviors and inconsistencies. Looking at distributor repo, you can see that this is not a simple topic, and that's why they don't support it yet.

Look the description in this issue:

https://github.com/10up/distributor/issues/112

And then see that there's a PR that never got merged https://github.com/10up/distributor/pull/446

So it seems to me that there might be some use cases we are not accounting for and simply syncing the post status in this way might break some editorial flows.

I can't really think of anything but I would feel more comfortable if we only covered what we know is an issue in the current workflow, which is trashing the post (and maybe restoring?). WDYT?

leogermani:
ah, and I pushed one little fix needed after all the code was put inside a class

adekbadek:
@leogermani – I've added a check so that only `publish` and `trash` statuses are synchronised.

leogermani (approved):
Thanks!

PR #46, opened 2024-02-12
fix: prevent coauthors capability check infinite loop

chickenn00dle:
Closes `0/1206331646877116/1206535636784422/f`

This fixes an issue where, when NP Network, Distributor, CAP, and NP Newsletters are active, sending a newsletters causes an infinite capability check which results in a fatal memory error.

The issue happens because we are creating a Distributor Post object to check a distributed posts linked status when filtering coauthors via the `get_coauthors` hook. But this same hook is eventually triggered in the instantiation of the Distributor Post object.

This PR fixes this by checking a posts linked status directly via post meta.

### Testing steps

1. Make sure you have a test site set up with Newspack Network, Distributor, Co-Authors Plus, and Newsletters, and mailchimp ESP setup
2. Create a new newsletter and send it to an audience while watching debug log
3. On trunk, you should see an memory fatal. On this branch you should not

Screenshot 2024-02-12 at 15 55 57

### DIFF
--- a/includes/distributor-customizations/class-authorship-filters.php
+++ b/includes/distributor-customizations/class-authorship-filters.php
@@ -44,13 +44,14 @@ class Authorship_Filters {
 		}
 
 		// We don't want to filter authors on admin, as it might break things.
-		if ( is_admin() || defined( 'REST_REQUEST' ) && REST_REQUEST ) {
+		if ( is_admin() ) {
 			return $coauthors;
 		}
 
 		// Only filter posts that are still linked to the original post.
-		$distributor_post = new DistributorPost( $post_id );
-		if ( ! $distributor_post->is_linked ) {
+		// We check the dt_unlinked post meta to determine linked status.
+		// See: https://github.com/10up/distributor/blob/1f180d74db804a057f7331ce62338e571dd73350/docs/post-meta.md?plain=1#L38.
+		if ( get_post_meta( $post_id, 'dt_unlinked', true ) ) {
 			return $coauthors;
 		}
 
--- a/includes/distributor-customizations/class-authorship-filters.php
+++ b/includes/distributor-customizations/class-authorship-filters.php
@@ -44,7 +44,7 @@ class Authorship_Filters {
 		}
 
 		// We don't want to filter authors on admin, as it might break things.
-		if ( is_admin() ) {
+		if ( is_admin() || defined( 'REST_REQUEST' ) && REST_REQUEST ) {
 			return $coauthors;
 		}
 


leogermani:
Nice catch!

But will this make the posts included in the newsletters to not display the filtered by lines?

Did you figure out why exactly this causes an OOM?

chickenn00dle:
> But will this make the posts included in the newsletters to not display the filtered by lines?

Ah. I'm not sure. TBH I'm not very familiar with co-authors plus and haven't really set it up to test properly. But will try setting up and confirming today.

> Did you figure out why exactly this causes an OOM?

It looks like there might be an infinite loop triggered when filtering coauthors via our network customization. It looks like in the process of creating the distributor post for the newsletter, co-authors plus triggers the `get_coauthors` hook which repeats the cycle.

claudiulodro:
That is a pretty common way of ending up with infinite loops (circular hooks). Good sleuthing!

chickenn00dle:
@leogermani, I've set up and tested this and it looks like distributed, coauthored posts still appear with the expected by lines. Here are the inserted posts from a test email sent to myself:

Screenshot 2024-02-12 at 17 40 05


I'm not sure if this covers all of the expected cases, or if this is exactly what you were referring to. Let me know if I'm off base.

chickenn00dle:
For informational purposes, after @leogermani pointed out an issue with guest authors with this approach, I did some more digging into the exact problem here.

1. When Newspack Network is active, we filter a posts coauthors on the `get_coauthors` hook. In the process we create a DistributorPost object here: https://github.com/Automattic/newspack-network/blob/8276b43e71a87adc47a10909f4187452c5c314e8/includes/distributor-customizations/class-authorship-filters.php#L52
2. The distribor post fetches the permalink in its constructor here: https://github.com/10up/distributor/blob/1f180d74db804a057f7331ce62338e571dd73350/includes/classes/DistributorPost.php#L198
3. Which eventually triggers the `user_has_cap` hook via a call to `current_user_can`: https://github.com/WordPress/WordPress/blob/53b99200d6000ff43effb3fd615cbe6a65d6b5e4/wp-includes/link-template.php#L133C1-L133C46
4. The CAP plugin uses this hook to allow guest authors to edit posts and in the process gets all coauthors: https://github.com/Automattic/Co-Authors-Plus/blob/68fc80d5795fd83cedd8fa7a51fe30b015caf0c4/template-tags.php#L65
5. This eventually calls the `get_coauthors` hook: https://github.com/Automattic/Co-Authors-Plus/blob/68fc80d5795fd83cedd8fa7a51fe30b015caf0c4/template-tags.php#L40

leogermani:
Yay! Thanks for catching this @chickenn00dle ! So good when these things are surfaced.

It seems the only reason we are using a `DistributorPost` object is to check the `is_linked` property. Let's find another wayto do this.

Ideally by using a method provided by Distributor. But if not possible, by reading the data directly (I guess it's a post meta)

leogermani (approved):
Phenomenal

PR #47, opened 2024-02-13
fix: dynamic class property deprecated warnings

chickenn00dle:
This PR just addresses some warnings flooding the logs:

Screenshot 2024-02-13 at 10 40 58

Screenshot 2024-02-13 at 10 40 40

### Testing instructions

Code review should be fine, but feel free to checkout this branch and confirm these warning no longer appear whenever network related events are triggered.

### DIFF
--- a/includes/hub/stores/class-abstract-event-log-item.php
+++ b/includes/hub/stores/class-abstract-event-log-item.php
@@ -57,6 +57,13 @@ abstract class Abstract_Event_Log_Item {
 	private $timestamp;
 
 	/**
+	 * The action name associated with the Event Log Item.
+	 *
+	 * @var string
+	 */
+	private $action_name;
+
+	/**
 	 * Creates an instance of an Event Log Item
 	 *
 	 * Do not create an object of this class directly. Instead, use the Event_Log Store class.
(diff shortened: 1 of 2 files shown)


includes/hub/stores/class-abstract-event-log-item.php
```diff
@@ -56,6 +56,13 @@ abstract class Abstract_Event_Log_Item {
 	 */
 	private $timestamp;
 
+	/**
+	 * The action name associated with the Event Log Item.
+	 *
+	 * @var string
+	 */
+	private $action_name;
```
leogermani: It looks like the `$action` attribute above is not being used. I guess it was renamed in the code, but not here. It looks like we can safely rename that one instead of creating a new one

leogermani (changes requested):

includes/hub/stores/class-abstract-event-log-item.php
chickenn00dle: Good catch! Updated in https://github.com/Automattic/newspack-network/pull/47/commits/991f008c9911b73a153e595945e42d6c1db6fc17

leogermani (approved):

PR #51, opened 2024-02-14
feat(backfill): WP CLI script to backfill reader_registered, donation_new events

adekbadek:
### All Submissions:

Adds a WP CLI script to backfill historical data. 

1. Switch `newspack-plugin` to `feat/tweaks-for-network-update` branch (https://github.com/Automattic/newspack-plugin/pull/2935)
2. Disable this plugin on both the hub and a node site, in order to build up historical data
3. On both sites (using different email addresses): 
    1. register a user, 
    2. make a donation
    3. make a recurring donation and the cancel the related subscription in WC
5. Activate the `newspack-network` plugin on both sites
6. Verify that the new user registrations and donations are not present in the Hub's event log
7. Run `wp newspack-network data-backfill reader_registered --start=2024-01-01 --end=2024-12-31` on the Hub site, observe the output the dry run and confirm nothing was added to the event log. 
    1. Do the same with `donation_new` as the argument instead of `reader_registered`.
    2. And then with `donation_subscription_cancelled` 
9. Try with `--verbose` flag for more output
10. Run `wp newspack-network data-backfill reader_registered --start=2024-01-01 --end=2024-12-31 --live` on the Hub site and observe the event was added to the event log. 
    1. Do the same with `donation_new` as the argument instead of `reader_registered`.
    2. And then with `donation_subscription_cancelled` 
12. Repeat the last three steps on the Node site, but instead of the event log look as the scheduled events in the Node Settings
13. Run in the live mode _again_ to ensure that duplicate events are not added to the event log or the scheduled events* (in the CLI output, observe it's noted that no events were processed and how many were skipped as duplicates)
14. After the scheduled event is transmitted, observe it was added to the Hub's event log
15. Run the script with `all` as action name argument and observe it has the same effect as running the above two scripts one after another

\* unless the webhook request queue has be cleared already, but the event log should handle deduplication itself anyway

### Other information:

### DIFF
--- a/includes/cli/class-data-backfill.php
+++ b/includes/cli/class-data-backfill.php
@@ -167,7 +167,9 @@ class Data_Backfill {
 		foreach ( $orders as $order ) {
 			$order_id = $order->get_id();
 			$order_data = \Newspack\Data_Events\Utils::get_order_data( $order_id );
-
+			if ( ! $order_data ) {
+				continue;
+			}
 			$timestamp = strtotime( $order->get_date_completed() );
 			if ( $live ) {
 				self::process_event_entity( $order_data, $timestamp, 'donation_new' );
@@ -215,7 +217,7 @@ class Data_Backfill {
 		foreach ( $subscriptions as $subscription ) {
 			$subscription_data = \Newspack\Data_Events\Utils::get_recurring_donation_data( $subscription );
 			if ( ! $subscription_data ) {
-				return;
+				continue;
 			}
 			$timestamp = strtotime( $subscription->get_date( 'cancelled' ) );
 			if ( $live ) {
--- a/includes/cli/class-data-backfill.php
+++ b/includes/cli/class-data-backfill.php
@@ -356,14 +356,28 @@ class Data_Backfill {
 		}
 		if ( $live ) {
 			WP_CLI::line( '' );
-			WP_CLI::success(
-				sprintf(
-					'Processed %d %s events, skipped %d as duplicates.',
-					self::$results[ $action ]['processed'],
-					$action,
-					self::$results[ $action ]['duplicate']
-				)
-			);
+			if ( $action === 'all' ) {
+				$processed = 0;
+				$duplicate = 0;
+				foreach ( array_values( self::$results ) as $key => $value ) {
+					if ( isset( $value['processed'] ) ) {
+						$processed += $value['processed'];
+					}
+					if ( isset( $value['duplicate'] ) ) {
+						$duplicate += $value['duplicate'];
+					}
+				}
+				WP_CLI::success( sprintf( 'Processed %d events, skipped %d as duplicates.', $processed, $duplicate ) );
+			} else {
+				WP_CLI::success(
+					sprintf(
+						'Processed %d %s events, skipped %d as duplicates.',
+						self::$results[ $action ]['processed'],
+						$action,
+						self::$results[ $action ]['duplicate']
+					)
+				);
+			}
 		}
 		WP_CLI::line( '' );
 	}
--- a/includes/cli/class-data-backfill.php
+++ b/includes/cli/class-data-backfill.php
@@ -130,7 +130,8 @@ class Data_Backfill {
 			];
 			if ( $live ) {
 				self::process_event_entity( $user_data, strtotime( $user->user_registered ), 'reader_registered' );
-			} elseif ( $verbose ) {
+			}
+			if ( $verbose ) {
 				WP_CLI::line( '👉 ' . sprintf( 'User %s (#%d) registered on %s.', $user->user_email, $user->ID, $user->user_registered ) );
 			}
 			if ( self::$progress ) {
@@ -165,7 +166,8 @@ class Data_Backfill {
 			$timestamp = strtotime( $order->get_date_completed() );
 			if ( $live ) {
 				self::process_event_entity( $order_data, $timestamp, 'donation_new' );
-			} elseif ( $verbose ) {
+			}
+			if ( $verbose ) {
 				WP_CLI::line( '👉 ' . sprintf( 'Order #%d completed on %s.', $order_id, gmdate( 'Y-m-d H:i:s', $timestamp ) ) );
 			}
 			if ( self::$progress ) {
@@ -210,7 +212,8 @@ class Data_Backfill {
 			$timestamp = strtotime( $subscription->get_date( 'cancelled' ) );
 			if ( $live ) {
 				self::process_event_entity( $subscription_data, $timestamp, 'donation_subscription_cancelled' );
-			} elseif ( $verbose ) {
+			}
+			if ( $verbose ) {
 				WP_CLI::line( '👉 ' . sprintf( 'Subscription #%d cancelled on %s.', $subscription->get_id(), gmdate( 'Y-m-d H:i:s', $timestamp ) ) );
 			}
 			if ( self::$progress ) {
--- a/includes/cli/class-data-backfill.php
+++ b/includes/cli/class-data-backfill.php
@@ -25,7 +25,7 @@ class Data_Backfill {
 	 *
 	 * @var class
 	 */
-	private static $progress = false;
+	public static $progress = false;
 
 	/**
 	 * Initialize this class and register hooks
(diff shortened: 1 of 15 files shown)


includes/cli/class-data-backfill.php
```diff
@@ -0,0 +1,362 @@
...
+		} else {
+			WP_CLI::line( 'Running on a Node site – will create webhook requests. This may result in duplicate requests if the request queue was cleared already, but the Hub will handle the duplicates.' );
+		}
+		WP_CLI::line( '' );
+
+		if ( ! method_exists( '\Newspack\Reader_Activation', 'get_reader_roles' ) ) {
+			WP_CLI::error( 'Incompatible Newspack plugin version.' );
+		}
+		if ( $live ) {
+			WP_CLI::line( '⚡️ Heads up! Running live, data will be updated.' );
```
leogermani, line 125: We could add a confirmation prompt here

includes/cli/class-data-backfill.php
```diff
@@ -0,0 +1,362 @@
...
+	 * Process reader_registered events.
+	 *
+	 * @param string $start The start date.
+	 * @param string $end The end date.
+	 * @param bool   $live Whether to run in live mode.
+	 * @param bool   $verbose Whether to output verbose information.
+	 */
+	private static function process_reader_registered( $start, $end, $live, $verbose ) {
+		// Get all users registered between $start and $end.
+		$users = get_users(
```
leogermani: We could filter out users with `newspack_remote_id` metadata. No need to backfill users that came from other sites. We want users that were originally registered on this site

includes/incoming-events/class-abstract-incoming-event.php
```diff
@@ -46,17 +46,33 @@ class Abstract_Incoming_Event {
...
 	/**
 	 * Constructs a new Incoming Event
 	 *
 	 * @param string       $site      The origin site URL.
 	 * @param array|object $data      The data for this event.
 	 * @param int          $timestamp The timestamp for this event.
+	 * @param string       $action_name The action name for this event.
 	 */
-	public function __construct( $site, $data, $timestamp ) {
+	public function __construct( $site, $data, $timestamp, $action_name = false ) {
```
leogermani: we shouldn't need this. See my comment on the suggested architecture

includes/cli/class-data-backfill.php
```diff
@@ -0,0 +1,362 @@
+<?php
+/**
+ * Data Backfill scripts.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network;
+
+use Automattic\Jetpack\VideoPress\Data;
```
leogermani: left over

leogermani (changes requested):
This is looking good, I left a couple of comments, but I wanted to suggest an architectural change so this is more aligned to the rest of the plugin.

Instead of having everything inside the Data_Backfill class, we should keep it just with the CLI command.

Then we can create a new namespace, something like `Newspack_Network\Backfillers`, where we can have one class per event, following the same nomenclature define in `Accepted_Actions` and which we use in other places.

The `Backfiller_Abstract_Class` could implement a method that is common to every event, what is currently the `process_event_entity`.

And each class need to implement its own `get_events` method, that will return an array with `Income_Event` objects (instances of their specific classes, and not the abstract one). The `process_event_entity` method would receive such array and process each item.

In the command callback, we would not need the switch case https://github.com/Automattic/newspack-network/pull/51/files#diff-2fbd8473ad6861f6fec58b5890525ac5092f375cbe554e54f604a9871e490ed8R324 as we would simply create an instance of the backfiller for the required action.

And we could fallback to the default class, which would simply return that this action is not backfillable (something like we do here -> https://github.com/Automattic/newspack-network/blob/bcb06155cb7d148f3dc735dbd88f566f8cf6f65c/includes/hub/stores/class-event-log.php#L52-L56

I hope this makes sense to you.

I think it has a number of advantages, not only to keep consistency with the rest of the plugin, but also to keep small classes and avoid this class to grow indefinitely. It will also be easier to implement new backfillers for new actions, as all you need to know is implement one class, no need to touch several places in the code.

Since all the work is done and this is basically moving code around, I'm happy to do it if you prefer to focus on other tasks. Let me know!

includes/cli/class-data-backfill.php
adekbadek: It already requires passing the `--live` flag explicitly, I don't see why it should be confirmed on top of that.

includes/cli/class-data-backfill.php
adekbadek: Fixed in 111a7b4fde706d7f6de42f2237ee619e8a36b477

includes/cli/class-data-backfill.php
adekbadek: Oh, great suggestion! Done in 111a7b4fde706d7f6de42f2237ee619e8a36b477

adekbadek:
Re: architectural change – sounds good, but for later. There's already another PR - https://github.com/Automattic/newspack-network/pull/62 - relying on this PR.

includes/cli/class-data-backfill.php
```diff
@@ -0,0 +1,365 @@
...
+				'date_created' => $start . '...' . $end,
+				'limit'        => -1,
+			]
+		);
+		if ( ! $verbose ) {
+			self::$progress = \WP_CLI\Utils\make_progress_bar( 'Processing orders', count( $orders ) );
+		}
+		foreach ( $orders as $order ) {
+			$order_id = $order->get_id();
+			$order_data = \Newspack\Data_Events\Utils::get_order_data( $order_id );
```
leogermani: we need to check if it is a donation related order. see my review on the newspack plugin's PR

includes/cli/class-data-backfill.php
```diff
@@ -0,0 +1,365 @@
...
+				'email'    => $user->user_email,
+				'metadata' => [
+					// 'current_page_url' is not saved, can't be backfilled.
+					'registration_method' => empty( $registration_method ) ? 'backfill-script' : $registration_method,
+				],
+			];
+			if ( $live ) {
+				self::process_event_entity( $user_data, strtotime( $user->user_registered ), 'reader_registered' );
+			}
+			if ( $verbose ) {
```
leogermani: apparently when you run `live`, the progress bar overrides the verbose output. Maybe only show progress bar if verbose is false

includes/cli/class-data-backfill.php
```diff
@@ -0,0 +1,365 @@
...
+						'value'   => $end,
+					],
+				],
+			]
+		);
+		if ( ! $verbose ) {
+			self::$progress = \WP_CLI\Utils\make_progress_bar( 'Processing subscriptions', count( $subscriptions ) );
+		}
+		foreach ( $subscriptions as $subscription ) {
+			$subscription_data = \Newspack\Data_Events\Utils::get_subscription_data( $subscription );
```
leogermani: again we need to check if it's a donation subscription here. See my comment on the main plugin PR

What happened to me was that I had a bunch of cancellations for non-donation subscriptions. My backfill data was then full of subscription_cancelled events with empy data (because the check was on the Utils function).

So the check should not be there, and be here instead, so we don't include subscription data at all.

leogermani (changes requested):

includes/cli/class-data-backfill.php
adekbadek: Addressed in the `newspack-plugin` PR

includes/cli/class-data-backfill.php
adekbadek: `--verbose` and progress bar are mutually exclusive, independent of `--live`. If the progress bar is rendered, logging anything would break the progress bar rendering. I haven't found a way to render the progress bar _and_ at the same time log individual lines.

includes/cli/class-data-backfill.php
adekbadek: Handled in 0153127e99ce5135ec0192fcc46e009a57bb4a84

includes/cli/class-data-backfill.php
leogermani: yesterday I thought `--live` was enforcing the progress bar. But today it's working as expected.. my mistake

leogermani:
When you run the command with `all`, the totals at the end is always zeros

Captura de tela de 2024-02-22 09-05-47

leogermani (changes requested):
Looking good

I left one comment about the checks in donation_subscription_cancelled and donation_new, but other than that it seems to be working fine!

Also left a minor comment about the command output

leogermani:
Oh, I added the comments to the [commit,](https://github.com/Automattic/newspack-network/commit/0153127e99ce5135ec0192fcc46e009a57bb4a84) not to this PR.

Let me add them here as well for visibility

We also need to check if the data is empty for the order in donation_new, otherwise we get events with empty data for non donation orders.

And I think these checks should trigger a `continue` instead of a `return`

adekbadek:
> When you run the command with all, the totals at the end is always zeros

Fixed in f9cc61e3eaee781b711fa164cb659163f36b8a0b

leogermani (approved):

leogermani:
@adekbadek I have performed a bunch of tests and all seems to be working fine, including Memberships.

I'll leave it as pre-approved to unblock you. If you could run a few more tests it would be nice. This will need to be tested with real data anyway.

And I hope you like the refactor :)

PR #49, opened 2024-02-13
feat: in-progress manual sync

leogermani:

PR #53, opened 2024-02-15
feat: add option to manually sync users 

laurelfulford:
This PR adds an option to manually sync a user's data to the network. If the user doesn't exist in the network yet, it will also create them. Unlike the automatic syncing, this option will also sync a user's role. 

## Testing instructions

1. From your Hub site, edit one of your users; it's fine to start with one that's already been synced across your network. Confirm about halfway down the screen, under the "Newspack Network Tools" header, there should now be an option to manually sync a user: 

image

2. Change the user's role, and click Save. 
3. After saving, scroll back to the Newspack Network section, and click the 'Sync user across network' button. The page should refresh, and give you a success message at the top:

image

4. Confirm that this creates a `network_manual_sync_user` action in your Hub's Event Log. 
5. On your Node site, confirm that the change has been made to the corresponding user, and that they have the new role. 
6. Create a new user on your Hub site, and edit them to add information (like a description, Facebook link, etc). Save.
7. After saving, on that new user's profile click the 'Sync user across network' button. 
8. Confirm that the new user is synced to your node sites with all of their populated information.
9. Repeat editing an existing user's role, saving, and manually syncing from one of your node sites. Confirm that the information is updated on the hub site.
10. Repeat creating a new user, saving, and manually syncing from one of your node sites. Confirm that the information is updated on the hub site.

Replaces #49 as the PR, just so I'm not putting words in Leo's mouth! 

See: 1206331646877116-as-1206522226618445


includes/class-user-manual-sync.php
```diff
@@ -0,0 +1,239 @@
...
+ * Class to watch the user for updates and trigger events
+ */
+class User_Manual_Sync {
+
+	/**
+	 * The metadata we're syncing.
+	 *
+	 * @var array
+	 */
+	public static $meta_to_sync = [
```
leogermani: Let's not duplicate it, but read from that class instead

includes/class-user-manual-sync.php
```diff
@@ -0,0 +1,239 @@
...
+		return $url;
+	}
+
+	/**
+	 * Add button to manually sync users to the Edit User Profile screen.
+	 *
+	 * @param WP_User $user The current WP_User object.
+	 */
+	public static function add_manual_sync_button( $user ) {
+		// TODO: should this option limited to specific roles?
```
leogermani: I don't think so

includes/class-user-manual-sync.php
```diff
@@ -0,0 +1,239 @@
...
+	 *
+	 * @param WP_User $user The current WP_User object.
+	 * @return string $url The wp-admin URL.
+	 */
+	public static function get_manual_sync_link( $user ) {
+		$url = add_query_arg(
+			array(
+				'user_id'  => \esc_attr( $user->ID ),
+				'action'   => 'np_network_manual_user_sync',
+				'uid'      => \esc_attr( $user->ID ),
```
leogermani: No need to create another parameter with the same value. You can simply read from `user_id` in you callback down there

includes/class-user-manual-sync.php
```diff
@@ -0,0 +1,239 @@
...
+	 * Create URL to manually sync user.
+	 *
+	 * @param WP_User $user The current WP_User object.
+	 * @return string $url The wp-admin URL.
+	 */
+	public static function get_manual_sync_link( $user ) {
+		$url = add_query_arg(
+			array(
+				'user_id'  => \esc_attr( $user->ID ),
+				'action'   => 'np_network_manual_user_sync',
```
leogermani: I would chose a more specific key, other than `action`, to avoid conflicting with other plugins or anything else

includes/class-user-manual-sync.php
```diff
@@ -0,0 +1,239 @@
...
+			}
+		}
+
+		if ( ! isset( $_GET['action'] ) || ! in_array( $_GET['action'], $actions, true ) ) {
+			return;
+		}
+
+		$action = \sanitize_text_field( \wp_unslash( $_GET['action'] ) );
+
+		if ( ! isset( $_GET['uid'] ) ) {
```
leogermani: you can change to user_id here

includes/class-user-manual-sync.php
```diff
@@ -0,0 +1,239 @@
...
+	 * Process the admin action request
+	 */
+	public static function process_admin_action() {
+		/** Allowed actions */
+		$actions = [
+			'np_network_manual_user_sync',
+		];
+
+		/** Add notice if admin action was successful. */
+		if ( isset( $_GET['update'] ) && in_array( $_GET['update'], $actions, true ) ) {
```
leogermani: I would also chose a more specific key here instead of `update`.

And a nitpck, since there is only one thing to check, no need to create an array and check in_array. You could simply do

` ... && $_GET['newspack_network_manual_sync_processed'] === 'success' `

includes/incoming-events/class-user-manually-synced.php
```diff
@@ -0,0 +1,91 @@
...
+
+		// Loop through user meta and update.
+		if ( isset( $data->meta ) ) {
+			foreach ( $data->meta as $meta_key => $meta_value ) {
+				Debugger::log( 'Manually syncing user meta: ' . $meta_key );
+				update_user_meta( $user->ID, $meta_key, $meta_value );
+			}
+		}
+
+		User_Utils::maybe_sideload_avatar( $user->ID, $data->meta, true );
```
leogermani: There was a recent fix that moved this line up inside the `if ( isset( $data->meta ) )`  check to avoid warnings

includes/class-user-manual-sync.php
laurelfulford: That totally makes sense! I updated the `$user_props` as well to grab from the same place.

includes/class-user-manual-sync.php
laurelfulford: Comment deleted! 😄

includes/class-user-manual-sync.php
laurelfulford: D'oh! Thanks Leo! This has been fixed, and all the `uid` references have been changed to `user_id`.

includes/class-user-manual-sync.php
laurelfulford: I updated these and the `update` ones to `np_network`-related values (`np_network_manual_sync_action` and `np_network_manual_sync_update`).

includes/incoming-events/class-user-manually-synced.php
laurelfulford: Oh thanks -- moved!

includes/class-user-manual-sync.php
laurelfulford: Done and done -- thank you! I got rid of the array entirely and just went with a single value 🙂

includes/class-user-manual-sync.php
laurelfulford: Done!

laurelfulford:
Thanks for the code feedback @leogermani -- you definitely caught some things I didn't consider (and a couple I did on autopilot that didn't really make any sense!!) 🙂 Everything should be resolved!

leogermani:
@laurelfulford This is almost working flawless. I've tested all flows.

Note that this not only works when you edit a user in the hub and the changes go to all Nodes. You can make the changes in any Node and it will be sent to all other sites in the network. And it works fine.

But for, when we do `set_role`, this is not working as expected for some reason. The role is not replaced. What happens is that the user ends up with 2 roles instead... Have you seen this happening with you? Remember how this happened during our call? It happens every time.

Not sure if there's another plugin interfering (I have nothing fancy) or what this is.

I experimented replacing the `->set_role()` call with:

```PHP
wp_update_user(
	[
		'ID'   => $user->ID,
		'role' => $new_role,
	]
);
```

and it worked fine.

leogermani (changes requested):
left a comment

laurelfulford:
Thanks @leogermani! 

That's so weird! I had tweaked the role code a bit [in this commit](https://github.com/Automattic/newspack-network/pull/53/commits/bfc958d32a7435b8ef455fafb93df43ad9116f3d#diff-ccce892f907c03a4a37ce5584a1875023255e21d08dae968c586e5da0b1a373bL53-R58), and that seemed to stop the doubled-up roles for me. But with you change it still works fine -- I kind wonder if I was missing a test case 😬 

That update was added in [c3abe1f](https://github.com/Automattic/newspack-network/pull/53/commits/c3abe1fff2bfde7e05b8d37f1009b02114ed2a64)!

leogermani (approved):

PR #56, opened 2024-02-15
feat: add hub bookmarks to nodes

chickenn00dle:
Closes 1206274567818686-as-1206445911335988 

This PR adds the bookmarks dropdown that was previously only available on the hub to all nodes in the network.

To do this, we add a new network event which is triggered whenever a node is saved or updated, which propagates some basic node data to all nodes in the network. The nodes store this data as an option and use it to render this dropdown.

Screenshot 2024-02-20 at 17 55 54

### Testing Instructions

1. Have a network with at least two nodes setup (in addition to the hub)
2. With these changes checked out, go to wp-admin in the hub site, and select a node to edit via Newspack Network > Nodes > YOUR NODE
3. Click update (no need to make any changes)
4. Go to the event log in the hub via Newspack Network > Event Log
5. Confirm a nodes synced event appears:
Screenshot 2024-02-20 at 17 58 27
6. Wait for the event to propagate to the nodes, then visit the wp-admin of a node site.
7. Confirm the dropdown menu is available and contains an entry for the hub as well as the other node in the network, but not for itself.
8. Do the same for the other node site.
9. Go back to the hub site and trash one of the nodes.
10. Confirm another nodes synced event in the event logs
11. Wait for the event to propagate
12. Go to wp-admin for the node that has not been removed from the network
13. Confirm the removed node is not present in the dropdown menu

Note: The removed node still has the dropdown and still thinks its part of the network. I think we should address cleaning up not only this dropdown, but all network settings for a removed node in a follow-up PR.

### DIFF
--- a/includes/class-data-listeners.php
+++ b/includes/class-data-listeners.php
@@ -36,6 +36,7 @@ class Data_Listeners {
 		Data_Events::register_listener( 'woocommerce_subscription_status_changed', 'newspack_node_subscription_changed', [ __CLASS__, 'item_changed' ] );
 		Data_Events::register_listener( 'woocommerce_order_status_changed', 'newspack_node_order_changed', [ __CLASS__, 'item_changed' ] );
 		Data_Events::register_listener( 'newspack_network_user_updated', 'network_user_updated', [ __CLASS__, 'user_updated' ] );
+		Data_Events::register_listener( 'newspack_network_nodes_synced', 'network_nodes_synced', [ __CLASS__, 'nodes_synced' ] );
 	}
 
 	/**
@@ -84,5 +85,15 @@ class Data_Listeners {
 	 */
 	public static function user_updated( $user_data ) {
 		return $user_data;
+	}
+
+	/**
+	 * Filters the nodes data for the event being triggered
+	 *
+	 * @param array $nodes_data The nodes data.
+	 * @return array
+	 */
+	public static function nodes_synced( $nodes_data ) {
+		return [ 'nodes_data' => $nodes_data ];
 	}
 }
--- a/includes/hub/admin/class-nodes-list.php
+++ b/includes/hub/admin/class-nodes-list.php
@@ -54,7 +54,7 @@ class Nodes_List {
 				function ( $bookmark ) {
 					return sprintf( '<a href="%s" target="_blank">%s</a>', esc_url( $bookmark['url'] ), esc_html( $bookmark['label'] ) );
 				},
-				Node::get_bookmarks( $node->get_url() )
+				$node->get_bookmarks()
 			);
 			$allowed_tags = [
 				'a' => [
@@ -107,7 +107,7 @@ class Nodes_List {
 			];
 			$wp_admin_bar->add_node( $args );
 
-			foreach ( Node::get_bookmarks( $node->get_url() ) as $bookmark ) {
+			foreach ( $node->get_bookmarks() as $bookmark ) {
 				$sub_item_id = $item_id . '-' . sanitize_title( $bookmark['label'] );
 				$args        = [
 					'id'     => $sub_item_id,
--- a/includes/hub/admin/class-nodes-list.php
+++ b/includes/hub/admin/class-nodes-list.php
@@ -54,7 +54,7 @@ class Nodes_List {
 				function ( $bookmark ) {
 					return sprintf( '<a href="%s" target="_blank">%s</a>', esc_url( $bookmark['url'] ), esc_html( $bookmark['label'] ) );
 				},
-				$node->get_bookmarks()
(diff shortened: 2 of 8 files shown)


includes/class-data-listeners.php
```diff
@@ -85,4 +86,14 @@ public static function item_changed( $item_id, $status_from, $status_to, $item )
...
 	}
+
+	/**
+	 * Filters the nodes data for the event being triggered
+	 *
+	 * @param array $nodes_data The nodes data.
+	 * @return array
+	 */
+	public static function nodes_synced( $nodes_data ) {
+		return [ 'nodes_data' => $nodes_data ];
```
chickenn00dle, line 97: I ran into some issues treating `$data` as an array in the incoming events class so had to massage this here to work with the nodes data array.

includes/hub/admin/class-nodes-list.php
```diff
@@ -54,7 +54,7 @@ public static function posts_columns_values( $column, $post_id ) {
 				function ( $bookmark ) {
 					return sprintf( '<a href="%s" target="_blank">%s</a>', esc_url( $bookmark['url'] ), esc_html( $bookmark['label'] ) );
 				},
-				$node->get_bookmarks()
+				Node::get_bookmarks( $node->get_url() )
```
leogermani: See my other comment, I think we could keep this method as it is

includes/hub/class-node.php
```diff
@@ -102,11 +106,12 @@ public function decrypt_message( $message, $nonce ) {
 	/**
 	 * Gets a collection of bookmarks for this Node
 	 *
+	 * @param  string $url The URL of the Node.
 	 * @return array
 	 */
-	public function get_bookmarks() {
+	public static function get_bookmarks( $url ) {
```
leogermani: This is kind of a nit, but can I ask that instead of making this method static, and then having to change its usage, can we create another static method that could be used in the new places

Then this method could be a shortcode for the other static method. It would just call it passing the url.

The new static method could live in this class, or maybe it makes sense for it to live in the Nodes (plural) class

includes/hub/class-nodes.php
```diff
@@ -28,6 +27,7 @@ class Nodes {
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_post_type' ] );
 		add_action( 'save_post', [ __CLASS__, 'save_post' ] );
+		add_action( 'save_post_' . self::POST_TYPE_SLUG, [ __CLASS__, 'sync_nodes' ] );
```
leogermani: Pleas use the `newspack_network_node_saved` action. We have had race condition issues around this https://github.com/Automattic/newspack-network/pull/57

unrelated to this PR, but maybe as a safeguard we should make the callback that saves the nodes run on an higher priority (less then 10) here:

https://github.com/Automattic/newspack-network/blob/a9230d9e20cb9f59c78b56e6125d047b4be1386d/includes/hub/class-nodes.php#L30

leogermani (changes requested):
Beautiful! Working great.

Left a couple of minor comments but it's great to see more people creating events :)

includes/hub/class-node.php
chickenn00dle: Sure thing! Done in https://github.com/Automattic/newspack-network/pull/56/commits/9ac97753700b7d8298682b8b37a6d551c9e1e49c

includes/hub/class-nodes.php
chickenn00dle: Sure! One issue with using this action though is when a node is trashed from the edit posts page vs. the edit single post page. In this case the nonce is not present and so the sync event won't trigger. 

Is it possible to inject the nonce field to the edit posts page? If not, I guess we can either replace the trash and restore URLs to include our nonces, or simply remove the option to trash from the edit posts page altogether.

I'd love your thoughts on this before I make any changes here.

leogermani: I think you could listen to an additional hook to trigger this then

for example you listen to `newspack_network_node_saved` and also to `trashed_post`, calling the same callback.

chickenn00dle: > for example you listen to newspack_network_node_saved and also to trashed_post, calling the same callback.

This works too! I wasn't sure if we wanted to avoid using the standard wordpress post hooks to avoid running into similar issues in the future. In any case I made the change in https://github.com/Automattic/newspack-network/pull/56/commits/1e590d1458290550fb8e45cc40e445bbeffca0a6.

I also added some np network specific hooks for trashing and untrashing nodes in case we need to do more than just sync nodes in the future. This might be overkill though, so if you'd like I can remove these and just hook directly on the `un/trashed_post` hooks.

leogermani (approved):

PR #58, opened 2024-02-19
Feat/better config UX

adekbadek:
### All Submissions:

Adds better and less error-prone UX around adding nodes. 

See 1205909204837811-as-1206321490407427/f

1. Start with a new node or unlink an existing node*
2. Create a new Node in the Hub's Nodes view
3. Click "Link the site" and observe a new browser window opening with the Node site's admin (or login page)

image

4. Click "Save Changes" on the Node site admin panel to link the sites
5. Click the link from the Hub again and observe a notice saying "This site is already linked to this Hub.":

image

\* `wp option delete newspack_node_hub_url newspack_node_secret_key newspack_network_site_role`

### Other information:

### DIFF
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -191,12 +191,4 @@ class Admin {
 	 */
 	public static function enqueue_scripts() {
 	}
-
-	/**
-	 * Is updating the Node settings from URL?
-	 */
-	public static function is_updating_from_url() {
-		$action = filter_input( INPUT_GET, 'action', FILTER_SANITIZE_SPECIAL_CHARS );
-		return $action === self::LINK_ACTION_NAME;
-	}
 }
--- a/includes/class-admin.php
+++ b/includes/class-admin.php
@@ -21,6 +21,11 @@ class Admin {
 	const SETTINGS_SECTION = 'newspack_network_settings';
 
 	/**
+	 * The action name for the link-site functionality.
+	 */
+	const LINK_ACTION_NAME = 'newspack-network-link-site';
+
+	/**
 	 * Runs the initialization.
 	 */
 	public static function init() {
@@ -185,5 +190,13 @@ class Admin {
 	 * @return void
 	 */
 	public static function enqueue_scripts() {
+	}
+
+	/**
+	 * Is updating the Node settings from URL?
+	 */
+	public static function is_updating_from_url() {
+		$action = filter_input( INPUT_GET, 'action', FILTER_SANITIZE_SPECIAL_CHARS );
+		return $action === self::LINK_ACTION_NAME;
 	}
 }
--- a/includes/class-initializer.php
+++ b/includes/class-initializer.php
@@ -29,8 +29,11 @@ class Initializer {
 			Hub\Newspack_Ads_GAM::init();
 		}
 
-		if ( Site_Role::is_node() ) {
+		// Allow to access node settings before the site has a role, so it can be set via URL.
+		if ( Site_Role::is_node() || ! Site_Role::get() ) {
 			Node\Settings::init();
+		}
+		if ( Site_Role::is_node() ) {
 			if ( Node\Settings::get_hub_url() ) {
 				Node\Webhook::init();
 				Node\Pulling::init();
--- a/includes/class-initializer.php
+++ b/includes/class-initializer.php
@@ -27,6 +27,7 @@ class Initializer {
 			Hub\Database\Subscriptions::init();
 			Hub\Database\Orders::init();
 			Hub\Newspack_Ads_GAM::init();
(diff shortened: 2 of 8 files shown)


includes/class-initializer.php
```diff
@@ -29,8 +29,11 @@ public static function init() {
 			Hub\Newspack_Ads_GAM::init();
 		}
 
-		if ( Site_Role::is_node() ) {
+		// Allow to access node settings before the site has a role, so it can be set via URL.
+		if ( Site_Role::is_node() || ! Site_Role::get() ) {
```
leogermani, line 34: With this change the "Node settings" menu is visible even before I define the site's role. Don't think it's ideal;

WDYT to process the incoming link in the "Site Role" settings page instead?

includes/class-admin.php
```diff
@@ -186,4 +191,12 @@ public static function admin_init() {
 	 */
 	public static function enqueue_scripts() {
 	}
+
+	/**
+	 * Is updating the Node settings from URL?
+	 */
+	public static function is_updating_from_url() {
```
leogermani: Can this live in the class that will actually process the request instead?

includes/node/class-settings.php
```diff
@@ -205,6 +214,72 @@ public static function sanitize_secret_key( $value ) {
...
+	public static function linking_interface_notice() {
+		if ( ! Admin::is_updating_from_url() ) {
+			return;
+		}
+		$referrer = self::get_referrer();
+		if ( isset( $_REQUEST['settings-updated'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			// Remove the params from URL - settings were just saved.
+			if ( $referrer ) {
+				if ( ! \Newspack_Network\Site_Role::get() ) {
+					update_option( \Newspack_Network\Site_Role::OPTION_NAME, \Newspack_Network\Site_Role::NODE_ROLE );
```
leogermani: Nit, but a `\Newspack_Network\Site_Role::set_as_node()` method would look prettier

includes/hub/class-nodes.php
```diff
@@ -187,14 +187,31 @@ public static function nodes_dropdown( $current_node = '', $name = 'node_id', $e
...
+	 *
+	 * @param WP_Post $node_post The node post.
+	 */
+	private static function get_site_link_url( $node_post ) {
+		$node_url   = get_post_meta( $node_post->ID, 'node-url', true );
+		$secret_key = get_post_meta( $node_post->ID, 'secret-key', true );
+		return add_query_arg(
+			[
+				'page'       => \Newspack_Network\Node\Settings::PAGE_SLUG,
+				'secret_key' => $secret_key,
```
leogermani: maybe not a huge deal, but it's not best practice to share a secret key like this.

We should create a temporary key, that expires within minutes, that will allow the node to make a request to the Hub to fetch the key

adekbadek: I'm not sure this actually would increase security. The secret key would end up in a request URL anyway, and if someone has access to server logs they'd see the URL param from any request. And I don't think the risk warrants the added complexity.

includes/class-initializer.php
adekbadek: Fixed in faab3894dd4e461164165ea7d6a054327355054b. Moving the incoming link processing to "Site Role" would not allow the code to leverage the existing settings-updating code.

includes/class-admin.php
adekbadek: Done

includes/node/class-settings.php
adekbadek: Done

includes/hub/class-nodes.php
leogermani: Not exactly. The secret key would be in the body of a POST response, encrypted by https. no middle man can see it and no risk of this URL be in the browser history, or the user accidentally sharing it.

I don't think we need to be radical and 100% secure and follow all security recommendations, if we were to do that we would probably use a pair of private/public keys for each site (this would increase complexity) or use Application passwords in both ends (which would also increase complexity)... but having the secret key to be exposed like that in the URL I think is going to radical in the opposite direction

adekbadek: Query string [is also encrypted by HTTPS](https://blog.httpwatch.com/2009/02/20/how-secure-are-query-strings-over-https/), but you're right that the secret key would be persisted in browser history. Still a pretty elaborate attack scenario, but I agree it's not ideal.

adekbadek: (improved security add in https://github.com/Automattic/newspack-network/pull/64 by @leogermani)

leogermani (approved):

PR #61, opened 2024-02-20
Feat/custom esp meta fields

leogermani:
This PR adds an option to select which contact metadata fields will be synced to the ESP, by filtering Newspack's method that normalizes the contact.

This is useful for a network because Mailchimp has a very strict limit in the number of contact metadata fields you can have. If each site in the network was to sync every field with a different prefix, this can quickly reach that limit and some sites won't be able to sync their data.

## Testing the option 

* Make sure tests pass
* Checkout https://github.com/Automattic/newspack-plugin/pull/2940
* In the Hub, go to Newspack Network > ESP Metadata Sync settings

image

* Confirm that when you change the select to "Custom", you see a list of metadata fields to select:

image

* Make a selection and save changes. Do it a couple of times and make sure the settings persist as expected
* Confirm that you see a  new event in the Event Log for each change to the option

image

* Update a Node and make sure the option was persisted in the node as well:

```
get_option( 'newspack_network_esp_metadata' );
```

## Testing the fields

* In two sites, configure different prefixes in Newspack > Reader Revenue > RAS > Advanced
* Perform some actions that will trigger user sync, for example registration or buying a product
* Confirm in the ESP that only the selected fields were synced

## Copy

Please review all the copy, menu item labels, etc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  newspack: newspack/newspack@1.4.4
+  newspack: newspack/newspack@1.5.3
 
 commands:
   checkout_code:
--- a/bin/install-wp-tests.sh
+++ b/bin/install-wp-tests.sh
@@ -130,7 +130,7 @@ recreate_db() {
 	shopt -s nocasematch
 	if [[ $1 =~ ^(y|yes)$ ]]
 	then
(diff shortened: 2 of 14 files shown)


dkoo:
As discussed with the rest of @Automattic/newspack-product, this PR is no longer needed as we've moved the options for setting which metadata fields to sync to the main plugin in https://github.com/Automattic/newspack-plugin/pull/2989. This does mean that the fields need to be configured on each hub & node site in a network, but we can consider a Network-specific feature to propagate settings in the future.

PR #63, opened 2024-02-21
fix: memberships sync

adekbadek:
### All Submissions:

Fixes the memberships sync (#40). If a membership is created, it means that it already exists on a remote site (it was synced to the site running the sync). 

This can't actually be tested until memberships backfill (https://github.com/Automattic/newspack-network/pull/62) is in place.

### Other information:

### DIFF
--- a/includes/incoming-events/class-woocommerce-membership-updated.php
+++ b/includes/incoming-events/class-woocommerce-membership-updated.php
@@ -49,7 +49,7 @@ class Woocommerce_Membership_Updated extends Abstract_Incoming_Event {
 			return;
 		}
 
-		if ( ! class_exists( 'WC_Memberships' ) ) {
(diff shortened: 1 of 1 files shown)


leogermani (commented):
The goal is to prevent the metadata from being added when running the action a second time?

I think we need to add some safeguards there, because $user_membership can be a WP_Error object

Maybe add a `$is_creating` variable and only update the meta then?

adekbadek:
> The goal is to prevent the metadata from being added when running the action a second time?

Not exactly, the goal is to **only** add the metadata if the membership is a "cloned" one. 

> I think we need to add some safeguards there, because $user_membership can be a WP_Error object

The return type is `\WC_Memberships_User_Membership|null`, so I've added explicit `null` handling in 5b093a0cd2071878ff651bea09b36a943f0ea5e4.

Also, updated the safeguard against inactive plugin – as we've learned recently (https://github.com/Automattic/newspack-newsletters/pull/1437#discussion_r1494491359) it's best to directly check for the functions used.

leogermani (approved):

PR #64, opened 2024-02-21
feat: better security on automatic connection

leogermani:
Refactors the connection link and makes the Node request the secret key from the Hub using a (real one time) nonce.

## Testing

Same as https://github.com/Automattic/newspack-network/pull/58. Follow the flow and confirm it works.

Edit `request_secret_key` method and force it to return a WP_Error so you can test the error notice on the UI

Check that the button to "Go back to Hub" works.

Check copy


adekbadek (approved):
Thank you! This feature is better now. Added some small tweaks/formatting in 1de9b198957c1e1581c365f6c34f3ea410194416.

PR #67, opened 2024-02-22
feat(node): added wp cli command - process pending requests

jaredrethman:
### All Submissions:

This PR introduces a WP CLI command for processing `np_webhook_request`s with post-meta status in `pending` state. 

1. Ensure a Hub and Node relationship exists and is configured correctly.
2. Node: Add incorrect "Hub URL" - WP Admin > Newspack Network > Node Settings
3. Node: Perform individual (as many as you can, the more the better) user related updates i.e. update; name, lastname etc.
4. Node: Add correct "Hub URL" - WP Admin > Newspack Network > Node Settings
5. Open terminal and run `wp newspack-network process-webhooks` to process and delete pending requests

### Other information:

### DIFF
--- a/includes/node/class-webhook.php
+++ b/includes/node/class-webhook.php
@@ -41,8 +41,8 @@ class Webhook {
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_endpoint' ] );
 		add_filter( 'newspack_webhooks_request_body', [ __CLASS__, 'filter_webhook_body' ], 10, 2 );
-		if(defined('WP_CLI') && WP_CLI){
-			WP_CLI::add_command('np-network process', [__CLASS__, 'cli_process']);
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			WP_CLI::add_command( 'np-network process', [ __CLASS__, 'cli_process' ] );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


includes/node/class-webhook.php
```diff
@@ -104,4 +110,98 @@ public static function sign( $data, $nonce, $secret_key = null ) {
...
+			++$processed_count;
+			$progress->tick();
+		}
+
+		$progress->finish();
+
+		if ( ! empty( $unprocessed_request_ids ) ) {
+			WP_CLI::warning( "The following requests could not be deleted:\n" . wp_json_encode( $unprocessed_request_ids ) );
+		}
+		WP_CLI::success( "Successfully processed {$processed_count}/{$requests_count} pending requests." );
```
chickenn00dle: We probably want to do an additional check to confirm we've actually processed any requests before outputting the success message. I'd say if processed_count is 0 we should even output an error message. 

For example if I run this without fixing my hub URL the output looks like this:
Screenshot 2024-02-22 at 16 04 11

includes/node/class-webhook.php
```diff
@@ -104,4 +110,98 @@ public static function sign( $data, $nonce, $secret_key = null ) {
...
+				}
+			}
+			++$processed_count;
+			$progress->tick();
+		}
+
+		$progress->finish();
+
+		if ( ! empty( $unprocessed_request_ids ) ) {
+			WP_CLI::warning( "The following requests could not be deleted:\n" . wp_json_encode( $unprocessed_request_ids ) );
```
chickenn00dle: I think this message is meant to say '...could not be processed' vs. '...could not be deleted' considering we already delete warnings above.

chickenn00dle (changes requested):
Works great @jaredrethman! Nice work 🎉 

I left two small comments regarding output, then I think this is good to go.

includes/node/class-webhook.php
jaredrethman: Have converted `warning` to `error` and added return after. Great suggestion!

includes/node/class-webhook.php
jaredrethman: Right you are, fixed, nice catch! 🤦

includes/node/class-webhook.php
```diff
@@ -104,4 +110,99 @@ public static function sign( $data, $nonce, $secret_key = null ) {
...
+			}
+			++$processed_count;
+			$progress->tick();
+		}
+
+		$progress->finish();
+
+		if ( ! empty( $unprocessed_request_ids ) ) {
+			WP_CLI::error( "The following requests could not be processed:\n" . wp_json_encode( $unprocessed_request_ids ) );
+			return;
```
chickenn00dle, line 222: Nit. Feel free to ignore. I think we were okay with the warning here. For large batches of events where some fail and some succeed, I would expect to see information about what failed and what succeeded.

Thinking about this some more, something like the following would be ideal IMO:

success - all events processed
warning - some events processed, some events not processed
error - no events processed

chickenn00dle (approved):
Thanks for the quick turnaround! I have one small nit, but since the current messaging works and I don't wanna hold this up, gonna go ahead and approve.

includes/node/class-webhook.php
jaredrethman: Some solid suggestions there @chickenn00dle! I have implemented your suggestions, capture last error (i.e. the one that caused the error to be captured) and added some additional comments/docs etc 36049d408ed5d461f43049cbf192678a2d2b8ba3, all-in-all your suggestion adds a lot of value to the command, thank you 🤜 🤛  

## Success:
Screenshot 2024-02-22 at 17 37 05

## Warning: (hard to test)
Screenshot 2024-02-22 at 17 39 10

## Errors:
Screenshot 2024-02-22 at 17 40 29

includes/node/class-webhook.php
```diff
@@ -38,6 +41,9 @@ class Webhook {
 	public static function init() {
 		add_action( 'init', [ __CLASS__, 'register_endpoint' ] );
 		add_filter( 'newspack_webhooks_request_body', [ __CLASS__, 'filter_webhook_body' ], 10, 2 );
+		if ( defined( 'WP_CLI' ) && WP_CLI ) {
+			WP_CLI::add_command( 'np-network process', [ __CLASS__, 'cli_process' ] );
```
adekbadek: We use the full plugin slug as the WP CLI command base elsewhere in code. Also, the command name should be more descriptive:

 ```suggestion
				WP_CLI::add_command( 'newspack-network process-webhooks', [ __CLASS__, 'cli_process' ] );
```

adekbadek (changes requested):

includes/node/class-webhook.php
jaredrethman: Updated with your suggestions.

adekbadek (approved):

PR #72, opened 2024-02-27
Fix/backfill duplicate handling

adekbadek:
### All Submissions:

Fixes two issues:

1. If and event is not persisted in the event log, it should not be processed. 
2. When backfilling `reader_registration` events, disregard `current_page_url` metadata.

By accident this branch also contains #71, but it won't hurt to add that to the `alpha` release.

0. On `alpha`,
1. Trigger a new `reader_registered` event by registering as a new reader on the site
4. Observe in the Event Log that the event has a `metadata.current_page_url` key in the `data` property
5. Run the backfill for `reader_registered` events (e.g. `wp newspack-network data-backfill reader_registered --start=2024-02-26 --end=2024-12-31 --verbose`)
6. Observe a new `reader_registered` event added, even though a `reader_registered` for this reader is already present
7. Observe that the new event does not have the `metadata.current_page_url` key
8. Switch to this branch, repeat the testing steps, observe the duplicate event is not added 

### Other information:

### DIFF
--- a/includes/cli/backfillers/class-reader-registered.php
+++ b/includes/cli/backfillers/class-reader-registered.php
@@ -66,7 +66,6 @@ class Reader_Registered extends Abstract_Backfiller {
 
 			$incoming_event = new \Newspack_Network\Incoming_Events\Reader_Registered( get_bloginfo( 'url' ), $user_data, strtotime( $user->user_registered ) );
 			$events[] = $incoming_event;
-
 		}
 
 		return $events;
--- a/includes/cli/backfillers/class-reader-registered.php
+++ b/includes/cli/backfillers/class-reader-registered.php
@@ -53,6 +53,22 @@ class Reader_Registered extends Abstract_Backfiller {
 
 		$events = [];
 
+		// Disregard `current_page_url` metadata when checking for duplicates of reader registrations.
+		// This metadata is unavailable for backfilled events, so it's not possible to check for duplicates with it.
+		add_filter(
+			'newspack_network_event_log_get_args',
+			function( $args ) {
+				if ( $args['action_name'] === 'reader_registered' && isset( $args['data'] ) ) {
+					$data = json_decode( $args['data'], true );
+					if ( isset( $data['metadata']['current_page_url'] ) ) {
+						unset( $data['metadata']['current_page_url'] );
+					}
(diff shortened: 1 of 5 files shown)


chickenn00dle (changes requested):
Tested and looks good for the most part. I am running into one issue while testing though. With these changes checked out, I see that all events will trigger exactly one duplicate before being ignore by the data backfill script.

For example, Here I have created 3 users (test4, test5, and test6@example.com), and each have triggered an initial reader_registered event:

Screenshot 2024-02-27 at 10 28 14

When I run the backfill script the first time, each one triggers an additional reader_registered event:

Screenshot 2024-02-27 at 10 30 12

It isn't until the second time I run this command that duplicates are actually ignored:

Screenshot 2024-02-27 at 10 30 54

Is this expected?

adekbadek:
Not expected, you've uncovered a shortcoming of this solution. It assumes the only metadata missing will be `current_page_url`, but in your example there's also a `referer` metadata key. The solution should be to disregard all `data`. 62fdf8d3755abcafe1d4ed601b35a908f635a2fd should fix it.

chickenn00dle (changes requested):
Hmm. I'm still running into an issue, although the behavior has slightly changed. This time, everything but the first of a batch of user registrations is being duplicated. For example, here is test7, 8, and 9. Running the backfill script creates duplicate events for test 8 and 9:
  
Screenshot 2024-02-27 at 10 57 01

adekbadek:
This is a bigger issue than I thought! The code is comparing timestamps. The _event_ will be saved a few milliseconds after the actual _registration_. Thus when looking for an event at time of the registration, it won't be found. Fixed in ee7be33e944b9fbc912404e83a45a4be12909043

chickenn00dle (approved):
Nice work @adekbadek!

This is working now so gonna give a thumbs up.

I did have one question/nit which is can we simply disregard/drop timestamp as well if the only relevant information we need to consider are email and date? Definitely not a blocker though and something we can revisit if necessary.

adekbadek:
Good suggestion! For registrations that makes sense, because there should be just one per email. The timestamp-range solution should also work for other events, though, and for these disregarding the timestamp altogether is not desirable. There might be many of them, with only the timestamp as differentiator (e.g. donations).

PR #80, opened 2024-03-01
fix(network-pass): handle multiple sync'd membership plans

adekbadek:
### All Submissions:

The network pass feature (#40) was written in a way to handle only one synchronised membership plan per site. If more than one membership plan had the same network pass id, only one of these would get the sync treatment. This PR fixed it, so multiple plans can have the same network pass on a site. 

1. Set up synchronised memberships as described in #40, but with a twist – assign the same Network ID to two membership plans on one of the sites
2. Test the feature as described in #40 – create a membership on one site, observe the memberships are set up for linked membership plans
3. Verify that the site with two synchronised membership plans did grant the test user _both_ memberships

### Other information:

### DIFF
--- a/includes/incoming-events/class-woocommerce-membership-updated.php
+++ b/includes/incoming-events/class-woocommerce-membership-updated.php
@@ -55,7 +55,7 @@ class Woocommerce_Membership_Updated extends Abstract_Incoming_Event {
 
 		global $wpdb;
 
-		$local_plan_id = $wpdb->get_var( // phpcs:ignore
+		$local_plan_ids_query_results = $wpdb->get_results( // phpcs:ignore
 			$wpdb->prepare(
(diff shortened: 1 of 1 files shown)


leogermani:
Does it make sense for more than one plan to share the same network id though? I'm not in the loop of the requirements, but thinking about how this feature was created, I would go the other way and add a check when you save the membership plan to make sure you don't allow two plans to have the same network id.

The "network id" was meant to be the ID of this particular plan in the Network, and not "the ID of the network".

What's the use case for having one membership plan in one site mapping to more than one plan in another site? if a user becomes a member in one plan in one site it would become a member in more than one plan in the other? I think the relationship should always be 1 plan mapping to 1 plan. If you have more than one plan, each plan will have a different Network ID.

Feel free to ignore this comment if it does make sense, just wanted to share what was the original intention there

leogermani (changes requested):
I think we should close this PR in favor of https://github.com/Automattic/newspack-network/pull/108

leogermani:
Closed in favor of https://github.com/Automattic/newspack-network/pull/108

PR #83, opened 2024-03-19
feat: add Network-specific custom metadata to ESP syncs

dkoo:
Using a filter introduced in https://github.com/Automattic/newspack-plugin/pull/2989, creates a class that lets us add custom contact metadata fields for sites using the Network plugin. The first and only field we're adding so far is `Network Registration Site`, which shows which site in the network the registration originated from.

I'm not 100% certain this field is needed because upon testing, I saw that the `Registration Page` field contains the URL of the site that originated the registration, and this field is not overwritten after the user is propagated throughout the network. But even if we decide not to keep the `Network Registration Site` field, the infrastructure for adding custom fields seems useful enough to retain.

To test:

1. Check out this branch and https://github.com/Automattic/newspack-plugin/pull/2989.
2. Visit **Newspack > Engagement > Reader Activation > Advanced Settings** and confirm you see the new `Network Registration Site` field as an option in the Metadata Field Settings section:

Screenshot 2024-03-19 at 4 13 43 PM

3. Leave it selected and register as a new reader on one of the sites in your network.
4. Confirm that the contact is synced with the new field to the connected ESP.
5. Visit the **Newspack Network > Event Log** in the Hub site and confirm that the `Network Registration Site` field is appended to the data that propagates the user throughout the network.

Screenshot 2024-03-19 at 4 17 30 PM

---
  - https://app.asana.com/0/0/1206844902666878

### DIFF
--- a/includes/class-esp-metadata-sync.php
+++ b/includes/class-esp-metadata-sync.php
@@ -50,8 +50,6 @@ class Esp_Metadata_Sync {
 			$metadata['network_registration_site'] = \esc_url( \get_site_url() );
 		}
 
-		error_log( print_r( $metadata, true ) );
-
 		return $metadata;
 	}
 }
--- a/includes/class-esp-metadata-sync.php
+++ b/includes/class-esp-metadata-sync.php
@@ -1,0 +1,57 @@
+<?php
+/**
+ * Newspack Hub ESP Metadata settings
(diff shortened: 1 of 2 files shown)


laurelfulford (approved):
This works as described 🙌

PR #85, opened 2024-03-27
feat: add Network-specific custom metadata to ESP syncs (#83)

dkoo:
Hotfix version of #83 for faster deployment.


chickenn00dle (approved):
Since this has already been tested and approved, I just confirmed the changes are in sync with https://github.com/Automattic/newspack-network/pull/83/files

PR #86, opened 2024-03-29
fix(event-log): only show users filter with environment constant

dkoo:
On production sites, the list of users in the event log filter can get very, very long and cause significant slowdowns when viewing the Event Log admin page. This change simply hides the Users filter on this page unless a `NEWSPACK_NETWORK_EVENT_LOG_SHOW_USERS_FILTER` constant is set in wp-config.php.

To test:

1. On a Hub site, visit **Newspack Network > Event Log** and confirm that the Users select dropdown is no longer shown.
2. Add `define( 'NEWSPACK_NETWORK_EVENT_LOG_SHOW_USERS_FILTER', true );` to wp-config.php and confirm that it reappears


jaredrethman (approved):
Code looks good + tested and works. Approved!

PR #91, opened 2024-04-10
feat: handle user deletion event on the network

adekbadek:
### All Submissions:

Adds user deletion handling.

(Note: this is a part of integrating the changes from the `data-integrity-improvements` branch (https://github.com/Automattic/newspack-network/pull/89))

1. Register a user on one of the sites, wait for the data to sync*
2. Delete the user on one of the sites, wait for the data to sync
3. Observe the user was deleted from all sites
4. Observe there is only one `network_user_deleted` event in the event log

\* can be sped up by running `wp newspack-network process-webhooks` on the node and then `wp newspack-network sync-all` on the other node sites

### Other information:


chickenn00dle (changes requested):
Hey @adekbadek, 

I'm running into issues testing this one. A couple things:

1. When I delete the propagated user from a node, I see the delete event, but the user is never deleted from the hub site. Here is what I see on the node:
Screenshot 2024-04-24 at 10 54 36

And once the event propagates, this is what I see on the hub:
Screenshot 2024-04-24 at 10 56 57

But the user is never deleted on the hub:
Screenshot 2024-04-24 at 10 57 09

2. Might be related to the above, but when a user is deleted, I am also seeing the user updated event trigger (pictured in the screenshots above). I would expect this event not to be needed once a user is deleted. I wonder if this also  has something to do with why the user persists on the hub even after the deletion event is processed?

adekbadek:
1. The code was handling only deletion of Network Readers, so if a reader deleted their account on a non-origin site, the account would not be deleted elsewhere. This was erroneous, fixed in 998b4fd53075b638c4312f291943ef24371ea0c3
2. Fixed the superfluous `network_user_updated` event triggering by deletion in 1d6ff1c63fa324c6ac1788ca75a7cee2a48e78ae

includes/incoming-events/class-user-deleted.php
```diff
@@ -0,0 +1,75 @@
...
+		if ( ! \Newspack\Reader_Activation::is_user_reader( $user ) ) {
+			Debugger::log( sprintf( 'User %s is not a reader, skipping deletion.', $email ) );
+			return;
+		}
+
+		/** Make sure `wp_delete_user()` is available. */
+		require_once ABSPATH . 'wp-admin/includes/user.php';
+
+		// Don't broadcast this deletion on the network.
+		add_filter( 'newspack_network_process_user_deleted', '__return_false' );
```
chickenn00dle, line 62: Just noting this results in conflicting messaging in the logs where the account is logged as skipped, but then is logged as deleted:

```
[25-Apr-2024 14:07:47 UTC] [3110] /Users/raz/Sites/newspack/network/includes/class-data-listeners.php:103 User deletion with email: raz+networksubscriber3@automattic.com was skipped due to filter use.
[25-Apr-2024 14:07:47 UTC] [3110] /Users/raz/Sites/newspack/network/includes/incoming-events/class-user-deleted.php:70 User raz+networksubscriber3@automattic.com deleted.
```

Since the final log indicates the deletion was successful, I think this is not a blocker.

chickenn00dle (approved):
Nice! This is working as expected now. I noticed some strangeness with the logs, but considering the primary functionality is working and deletions are ultimately being logged, I'm gonna go ahead and approve this.

PR #98, opened 2024-05-13
Assorted changes 2

adekbadek:
### All Submissions:

1. Tweaks to the memberships deduplication CLI command, added in #84
2. A fix to username generation, to ensure too-long usernames are not used
3. Adds handling of syncing of users who already exist on the other site, but have a non-synchronizable role 
4. Some refactoring

(Note: this is a part of integrating the changes from the `data-integrity-improvements` branch (https://github.com/Automattic/newspack-network/pull/89))

1. Test membership deduplication command, as described in #84, but instead of a `--dry-run` flag observe it will dry run by default, making any changes only with `--live` flag provided. This brings it in line with other CLI command in this repository. 
2. On one network site, using CLI (front-end won't let you!), create a Subscriber user with an email address that exceeds 60 characters. 
    1. Wait for the user to sync, or trigger a sync
    2. Observe that the user's username on the other network sites is based on the email address, but cut to not exceed 60 characters
4. Create two users on two different network sites, with the same email address, but different roles: 
    1. Create an Editor (a non-synchronizable role) using WP Admin or WP CLI
    2. Register using the Registration block (or the auth modal) on the second site
    1. Run the reader-registered-event backfill (`wp newspack-network data-backfill reader_registered --start=2020-01-01 --end=2024-12-31 --live`) from the second site
    3. Wait for the user to sync, or trigger a sync
    4. Observe that on the first site, the user has both roles now 

### Other information:

### DIFF
--- a/includes/hub/admin/class-event-log.php
+++ b/includes/hub/admin/class-event-log.php
@@ -8,13 +8,12 @@
 namespace Newspack_Network\Hub\Admin;
 
 use Newspack_Network\Admin as Network_Admin;
+use const Newspack_Network\constants\EVENT_LOG_PAGE_SLUG;
 
 /**
  * Class to handle the Event log admin page
  */
 class Event_Log {
-
-	const PAGE_SLUG = 'newspack-network-event-log';
 
 	/**
 	 * Runs the initialization.
@@ -30,7 +29,7 @@ class Event_Log {
 	 * @return void
 	 */
 	public static function add_admin_menu() {
-		Network_Admin::add_submenu_page( __( 'Event Log', 'newspack-network' ), self::PAGE_SLUG, [ __CLASS__, 'render_page' ] );
+		Network_Admin::add_submenu_page( __( 'Event Log', 'newspack-network' ), EVENT_LOG_PAGE_SLUG, [ __CLASS__, 'render_page' ] );
 	}
 
 	/**
@@ -39,7 +38,7 @@ class Event_Log {
 	 * @return void
 	 */
 	public static function admin_enqueue_scripts() {
-		$page_slug = Network_Admin::PAGE_SLUG . '_page_' . self::PAGE_SLUG;
+		$page_slug = Network_Admin::PAGE_SLUG . '_page_' . EVENT_LOG_PAGE_SLUG;
 		if ( get_current_screen()->id !== $page_slug ) {
 			return;
 		}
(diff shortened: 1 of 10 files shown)


includes/hub/admin/class-event-log.php
```diff
@@ -8,14 +8,13 @@
 namespace Newspack_Network\Hub\Admin;
 
 use Newspack_Network\Admin as Network_Admin;
+use const Newspack_Network\constants\EVENT_LOG_PAGE_SLUG;
```
dkoo, line 11: This new constant needs to be imported in `class-users.php`, too: https://github.com/Automattic/newspack-network/blob/assorted-changes-2/includes/class-users.php#L78

dkoo (changes requested):
I'm not sure if this is something wrong with my local Network sites' setup, but for steps 2 and 3, when the node site's users got synced to my hub site, they came in without any role.

Screenshot 2024-05-15 at 5 29 19 PM

Also getting a fatal error in the Users page—see below.

adekbadek:
> when the node site's users got synced to my hub site, they came in without any role.

That should not ever happen, can you please list the exact reproduction steps?

Also, I've realized the step 3 instructions were wrong (synced user has to register using RAS, not be created with CLI), so I've updated this step. I don't think this relates to the issue described above, though.

includes/hub/admin/class-event-log.php
adekbadek: Fixed in 8f91df9d45cebd74ce4f9ea3082386cc0f99ed0f, thanks!

adekbadek:
@dkoo – I've requested a re-review, but what I meant was pinging you about the reproductions steps, as mentioned in the comment above.

dkoo:
Got this working. Everything is working as described except for this step:

> Observe that the user's username on the other network sites is based on the email address, but cut to not exceed 60 characters

In my testing, the user with the really long email address got propagated across the network, but the username was set to match the username of the account on the original site (not the email address). This username string was already shorter than 60 characters because CLI wouldn't let me create a user with a username longer than that.

dkoo (approved):

adekbadek:
Ah, that's right – the `user_name` will be overridden if already set on the origin site:

https://github.com/Automattic/newspack-network/blob/3f9755ed5b75a280c0a0e23b6d91f319ad0b8fa4/includes/incoming-events/class-user-manually-synced.php#L59

PR #108, opened 2024-07-01
feat: ensure that membership plans have unique network id

leogermani:
One membership plan in one site should always map to one membership plan in the other sites.

This PRs make sure you don't allow users to set the same "network id" for more than one membership plan.

This is an alternative to https://github.com/Automattic/newspack-network/pull/80

## Testing

* Create a membership plan and assign a network id
* Confirm it saves the network id as you created it
* edit the membership plan and confirm the network id is kept
* Create another membership plan and try to assign the same network id
* confirm the network id is suffixed with `-2`
* Try to create a third one and confirm it's still unique
* Edit any of the membership plans and add an unique network id and make sure it does not get the `-2` suffix added when it shouldn't


chickenn00dle (approved):
LGTM 👍

PR #112, opened 2024-07-16
feat(subscriptions): limit network-sync'd subscriptions buying

adekbadek:
### All Submissions:

Restricts subscription buying if the user has already a subscription on another network site (if that subscription is tied to a synchronised membership plan).  

image

See 1200133283036252-as-1207352955560208/f

1. Create a granted-on-subscription memberships plan on two network sites, and make them synchronised (set same `Network ID` in the plan settings)
3. One one site, buy the subscription, which should grant you the plan
4. On the second site, log in/register using the same email address
5. On the second site, try to buy the membership-granting subscription – observe it's not allowed (see screenshot above)

### Other information:

---
  - https://app.asana.com/0/0/1207352955560208

### DIFF
--- a/includes/woocommerce-subscriptions/class-limiter.php
+++ b/includes/woocommerce-subscriptions/class-limiter.php
@@ -1,0 +1,124 @@
+<?php
+/**
+ * Newspack Network limiter for WooCommerce Subscriptions.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network\Woocommerce_Subscriptions;
+
+/**
+ * Handles limiting for WooCommerce Subscriptions - only one
+ */
+class Limiter {
+	/**
+	 * Cache for restriction checking results.
+	 *
+	 * @var array
+	 */
+	private static $cache = [];
+
+	/**
+	 * Initializer.
+	 */
+	public static function init() {
+		add_filter( 'woocommerce_subscription_is_purchasable', [ __CLASS__, 'restrict_network_subscriptions' ], 10, 2 );
+		add_filter( 'woocommerce_cart_product_cannot_be_purchased_message', [ __CLASS__, 'woocommerce_cart_product_cannot_be_purchased_message' ], 10, 2 );
+	}
+
+	/**
+	 * Restricts subscription purchasing from a network-synchronized plan to one.
+	 *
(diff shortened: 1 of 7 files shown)


includes/woocommerce-subscriptions/class-limiter.php
```diff
@@ -0,0 +1,124 @@
...
+	 * Filters the error message shown when a product can't be added to the cart.
+	 *
+	 * @param string     $message Message.
+	 * @param WC_Product $product_data Product data.
+	 *
+	 * @return string
+	 */
+	public static function woocommerce_cart_product_cannot_be_purchased_message( $message, $product_data ) {
+		if ( ! self::can_buy_subscription( $product_data ) ) {
+			return __( 'You can only purchase one subscription in this network at a time.', 'newspack-network' );
```
leogermani, line 120: Suggestion: Instead of just saying they can't buy the subscription, we could say that they already have that subscription, because they bought it in site X

leogermani (changes requested):
I understand the value of these new Data endpoints to audit and check the health of the sync across sites. 

But for the regular operations of the plugin we shouldn't rely on it.

In this case we are making a request to the Hub, that will potentially make a dozen of requests to all the other sites. It can take a while and requests could fail.

The way to do it, following what the plugin does with everything else, is to sync the information about what subscriptions/memberships the user has across all sites.

So when a user tries to buy a subscription, we can check if they already have it by looking at local data, no need for external requests.

One way to do it is to store a `user_meta` whenever a user buys a subscriptions, recording everything we need (site id, plan, id, membership network id, etc).

Then you can add this meta to the [list of `user_meta` we sync](https://github.com/Automattic/newspack-network/blob/106497c51a551471e35651683903e1cef42f6e1f/includes/class-user-update-watcher.php#L35) and we'll automatically have it available in all sites.

But looking at it I wonder if there isn't an even simpler solution. Because when a user buys a membership in one site, it gets propagated and they are granted the membership in all sites. So when they try to buy a subscription that will enable a membership can't we just check if they user already has that membership plan and halt there?

leogermani:
closing in favor of https://github.com/Automattic/newspack-network/pull/138

PR #115, opened 2024-07-18
feat: create data event to sync user

leogermani:
This PR does two things

Removes the sync of users from the `newspack_network_network_reader` and leaves it only for `newspack_network_new_network_reader`.

Instead of calling the main plugin's data events method directly, fires a custom Data Events API event and leave the handling of the event to the main plugin, since this is a RAS feature.

## Testing

* Checkout https://github.com/Automattic/newspack-plugin/pull/3259 and the `epic/consolidate-data-flows` on newspack-newsletters plugin
* Set up two sites in a network, one hub and one node
* Make sure both sites have RAS enabled and Sync to ESP option turned on (check if you need the `NEWSPACK_FORCE_ALLOW_ESP_SYNC` constant, I did.)
* Make sure you have an "Audience ID" or Master List configured
* In one of the sites, perform an action that will register a new user, like registering or donating
* Watch the logs for the other site, and wait for it to add the new user
* When the new user is processed and added, make sure you see `Executing action handlers for "network_new_reader".` in your logs
* Right after that, check for the `Normalizing contact data for reader ESP sync:` with all the contacts metadata, make sure `Network Registration Site` is present

### DIFF
--- a/includes/class-esp-metadata-sync.php
+++ b/includes/class-esp-metadata-sync.php
@@ -8,6 +8,7 @@
 namespace Newspack_Network;
 
 use Newspack_Network\Utils\Users as User_Utils;
+use Newspack\Data_Events;
 
 /**
  * Class to handle Node settings page
@@ -23,8 +24,20 @@ class Esp_Metadata_Sync {
 		\add_filter( 'newspack_ras_metadata_keys', [ __CLASS__, 'add_custom_metadata_fields' ] );
 		\add_filter( 'newspack_register_reader_metadata', [ __CLASS__, 'handle_custom_metadata_fields' ], 10, 2 );
 		\add_filter( 'newspack_data_events_reader_registered_metadata', [ __CLASS__, 'handle_custom_metadata_fields' ], 10, 2 );
-		\add_action( 'newspack_network_network_reader', [ __CLASS__, 'handle_custom_metadata_for_network_readers' ] );
-		\add_action( 'newspack_network_new_network_reader', [ __CLASS__, 'handle_custom_metadata_for_network_readers' ] );
+		\add_action( 'init', [ __CLASS__, 'register_listeners' ] );
+	}
+
+	/**
+	 * Register the listeners to the Newspack Data Events API
+	 *
+	 * @return void
+	 */
+	public static function register_listeners() {
+		if ( ! class_exists( 'Newspack\Data_Events' ) ) {
+			return;
+		}
+
+		Data_Events::register_listener( 'newspack_network_new_network_reader', 'network_new_reader', [ __CLASS__, 'new_reader_data_event' ] );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


adekbadek (commented):
Let's call site A the site where I'm registering as the new user, and site B the other site.
I do see `Executing action handlers for "network_new_reader".` in the site B logs only and `Normalizing contact data for reader ESP sync:` is site A logs only. Is this expected? From the testing instructions I assume both log lines should be visible in site B's logs.

dkoo (changes requested):
I'm running into an error on the site receiving the new network reader event. The normalized data is missing an email address:

```
[25-Jul-2024 22:16:13 UTC] [10980][NEWSPACK][Newspack\Newspack_Newsletters::normalize_contact_data]: {
    "email": "",
    "metadata": {
        "NP_Account": 31,
        "NP_Registration Date": "2024-07-25",
        "NP_Membership Status": "",
        "NP_Product Name": "",
        "NP_Network Registration Site": "https:\/\/node-site.test",
        "status": "subscribed"
    },
    "existing_contact_data": false
}
```

which causes the Mailchimp API request to update the contact to return a 400 response with an error `Search query string must be non-empty, try again`. I figured this out because I kept seeing `PHP Warning:  Undefined array key "exact_matches" in /newspack-repos/newspack-newsletters/includes/service-providers/mailchimp/class-newspack-newsletters-mailchimp.php on line 1584` in the log and saw that the contact doesn't get synced from the receiving site.

leogermani:
Thanks @adekbadek and @dkoo 

> From the testing instructions I assume both log lines should be visible in site B's logs.

That's correct. Both things should be visible on site B. Make sure RAS is enabled on site B, sync to ESP is turned on and that there is a Master List/Audience selected.

> The normalized data is missing an email address:

Fixed!

adekbadek:
> Both [log lines] should be visible on site B.

When testing again, it's still [the same](https://github.com/Automattic/newspack-network/pull/115#pullrequestreview-2191266180).

leogermani:
> > Both [log lines] should be visible on site B.
> 
> When testing again, it's still [the same](https://github.com/Automattic/newspack-network/pull/115#pullrequestreview-2191266180).

Can you see the contact beeing synced to the ESP? 

One thing I corrected in the testing instructions that I forgot to mention is that you also need to checkout newspack-newsletters on `epic/consolidate-data-flows`.

@dkoo Do you see the logs and the sync happening?

dkoo:
> @dkoo Do you see the logs and the sync happening?

No, I'm still seeing the same behavior. Did you forget to push the fix to the remote?

leogermani:
I thought I was going crazy for a moment, but the fix for this was mistakenly committed directly to the epic branch on the main plugin here -> https://github.com/Automattic/newspack-plugin/commit/efaa47fad6b36caae60938628357f3d310b3eb59

I guess with all the branches / plugins switches I thought I was pushing the fix to this branch, but it turns out to be in another plugin :) 

So you need to make sure to update your main plugin in the `epic/consolidate-data-flows` branch. sorry

dkoo (approved):
@leogermani thanks for clarifying! After updating the other plugins to the correct branches, I now see the `network_new_reader` event happening on site B (the one receiving the network reader) and I see the sync getting triggered there as well.

leogermani:
Thanks @dkoo 

FYI, https://github.com/Automattic/newspack-network/pull/115/commits/9c5d3f3cd02ba485473cfa2ec3a013dea1bbaeda is not a big deal, and should not make a difference in production. But for my local sites, the local URLs were not passing that validation and the event was being triggered with the wrong URL there. I think it's fine to leave it without that check because we are not doing anything dangerous with it

PR #122, opened 2024-08-07
Epic/subscription membership sync

leogermani:
This epic refactors the strategy to check for users' subscriptions across the network, as well as recovering the Woo centralized data based on sync events.

I did this on a series of chained PRs to make it easier to parse. They need to be reviewed and merged in order into the Epic branch. Last step is to merge this epic branch into trunk.

In this epic branch, we revert https://github.com/Automattic/newspack-network/pull/114, https://github.com/Automattic/newspack-network/commit/cb5bcb7be0235c72a21d89ffcdac6ae154a3fa15 and https://github.com/Automattic/newspack-network/commit/e23442525922b95d66455ff48b40b1cb85ed7eeb, and do a little fix to ensure the data is always parsed the same way (https://github.com/Automattic/newspack-network/pull/122/commits/9140a8ebe8accd71d1eba19a46f532f10a4d30fa)

### 1: https://github.com/Automattic/newspack-network/pull/123

This simply creates backfillers for the existing Woo events and creates a user meta where we store the user's network subscriptions.

Testing:

* On a site with existing Woo data, run `wp newspack-network data-backfill newspack_node_subscription_changed` and `wp newspack-network data-backfill newspack_node_order_changed`. Test it in `verbose` and normal modes
* Make sure the events are created correctly
* Propagate the events in the network and make sure the Central Woo dashboard looks right (use `wp newspack-network process-webhooks` in the Node to speed syncing up)
* Find a user that has a subscription in a site and take note of their email
* Find this user in another site in the network
* Get the usermeta: `get_user_meta( $user_id, '_newspack_network_subscriptions', true);`
* Confirm the user meta contains the subscriptions this user has in other sites in the network

###  2: https://github.com/Automattic/newspack-network/pull/124

This PR creates the `membership_plan_updated` event

Testing

* On a site with existing membership plans, run `wp newspack-network data-backfill newspack_network_membership_plan_updated` 
* Confirm events for all existing membership plans is created
* Propagate the events (either pulling on the Node site or using `process-webhooks` to send events to the hub)
* Check the option on the sites include all existing membership plans in the other sites `get_option( 'newspack_network_membership_plans' );`
* Edit a membership plan in one site
* Confirm it created a new `membership_plan_updated` event
* Propagate the event
* Make sure the option is updated in the target site

### https://github.com/Automattic/newspack-network/pull/134

The above PR brings back #114 .

[...]

PR #123, opened 2024-08-07
Add/subscriptions backfillers

leogermani:


adekbadek (approved):

includes/hub/stores/class-subscriptions.php
```diff
@@ -58,18 +58,22 @@ public static function persist( Subscription_Changed $subscription ) {
 
 		Debugger::log( 'Persisting subscription ' . $subscription_id );
 
-		$local_id          = self::get_local_id( $subscription );
-		$subscription_data = self::fetch_data_from_api( $subscription );
```
adekbadek, line 62: `$subscription_data` is removed here, but still used on line 86.

adekbadek (changes requested):

includes/hub/stores/class-subscriptions.php
leogermani: It seems I forgot to update that to add the products. Fixed in https://github.com/Automattic/newspack-network/pull/123/commits/dbc12f1f4a49955abcf544e5778eb9b3e68fa174

Also noticed that I forgot to delete the listeners from generic class after I moved them to woocommerce/events.php . Fixed in -> https://github.com/Automattic/newspack-network/pull/123/commits/2adaaea5d5603a186195885d0bea86db2fd536ba . This was certainly what was causing the [duplicate events](https://github.com/Automattic/newspack-network/pull/122#pullrequestreview-2253961847)

adekbadek (approved):

PR #125, opened 2024-08-07
feat: handle active subs in other sites when cancelling memberships

leogermani:


tests/unit-tests/test-subscription-integration.php
```diff
@@ -0,0 +1,432 @@
...
+		$this->assertEquals( [], $result );
+
+		$result = $method->invoke( null, self::$user_with_non_network_sub_on_1 );
+		$this->assertCount( 1, $result );
+		$this->assertArrayHasKey( 'http://site1', $result );
+		$this->assertCount( 1, $result['http://site1'] );
+		$this->assertArrayHasKey( 503, $result['http://site1'] );
+
+		$result = $method->invoke( null, self::$user_without_subs );
+		$this->assertEquals( [], $result );
```
adekbadek: These tests are a bit hard to read with the generic variable names. It'd be more readable if `$method` was called e.g. `$get_user_network_active_subscriptions_method` and `$result` e.g. `$user_network_active_subscriptions`.

adekbadek (approved):
Works as expected, but the `Membership is not cancelled…` note was appended three times to the membership, which seems excessive :)

image

tests/unit-tests/test-subscription-integration.php
leogermani: https://github.com/Automattic/newspack-network/pull/125/commits/682f694368e9f1471de184e06fdbf822bd6b1a07

leogermani:
> Works as expected, but the `Membership is not cancelled…` note was appended three times to the membership, which seems excessive :)
> 
> image

Could not replicate this in my tests...

adekbadek:
When I tested it now, the note was added twice, instead of thrice ¯\_(ツ)_/¯. Perhaps because of the two filters used with the `prevent_membership_expiration` callback? Anyway, this is not that big of an issue, and mitigation might add unnecessary complexity.

adekbadek (approved):

leogermani:
Closing in favor of https://github.com/Automattic/newspack-network/pull/135

PR #126, opened 2024-08-07
feat: limit subscription purchase for user with network membership

leogermani:

### DIFF
--- a/includes/class-initializer.php
+++ b/includes/class-initializer.php
@@ -58,6 +58,7 @@ class Initializer {
 		Woocommerce_Memberships\Admin::init();
 		Woocommerce_Memberships\Events::init();
 		Woocommerce_Memberships\Subscriptions_Integration::init();
+		Woocommerce_Memberships\Limit_Purchase::init();
 
 		Woocommerce\Events::init();
 
--- a/includes/woocommerce-memberships/class-limit-purchase.php
+++ b/includes/woocommerce-memberships/class-limit-purchase.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ * Newspack Network limiter for WooCommerce Subscriptions.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network\Woocommerce_Memberships;
+
+/**
+ * Handles limiting WooCommerce Subscriptions purchases
+ *
+ * If the subscription is tied to a membership, we check whether the user already has an active subscription that gives access to the same membership in another site in the network.
+ */
+class Limit_Purchase {
+
+	/**
+	 * Initializer.
+	 */
+	public static function init() {
+		add_filter( 'woocommerce_subscription_is_purchasable', [ __CLASS__, 'restrict_network_subscriptions' ], 10, 2 );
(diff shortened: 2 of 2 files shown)


adekbadek (changes requested):
Can't test successfully – the subscription is still purchasable on the node site. Tracing the code, I see that it relies on the `newspack_network_membership_plans` option being set on the node site. In my setup, this option is only added on the hub site.

leogermani:
> Can't test successfully – the subscription is still purchasable on the node site. Tracing the code, I see that it relies on the `newspack_network_membership_plans` option being set on the node site. In my setup, this option is only added on the hub site.

Maybe you tried to buy before the new plan was pulled to the Node?

I updated the testing instructions with more details to make sure the info about all Memberhip Plans are populated before you try to buy. It's basically what was tested on #124

adekbadek:
Ah, I see. I misread the instructions for #124 and thought the `newspack_network_membership_plans` option is expected on the node site only. 

After updating the membership plan on the node and synchronizing the events, the `newspack_network_membership_plans` option is present on both sites. But this is problematic, because it means a manual migration would be needed after these changes are deployed – the network-synced plans on all sites would have to be fake-updated to trigger the `newspack_network_membership_plans` events. 

Can the migration (populating `newspack_network_membership_plans` option) be done automatically?

adekbadek:
Note: handling of non-logged in users should also be added (with email-lookup), similar to https://github.com/Automattic/newspack-blocks/pull/1816.

leogermani:
> After updating the membership plan on the node and synchronizing the events, the newspack_network_membership_plans option is present on both sites. But this is problematic, because it means a manual migration would be needed after these changes are deployed – the network-synced plans on all sites would have to be fake-updated to trigger the newspack_network_membership_plans events.

Yes, a backfill of membership plans and also all the woo data will be necessary. The backfillers added here should address them all. But yes, it is a step we'll need to go through.

Maybe we can work on that idea to have a central backfiller command, from the Hub, that will remotely reach each node.. this would save some steps.

But in any case, I'm happy to test and run it all

leogermani:
@adekbadek here's an idea

We bring back https://github.com/Automattic/newspack-network/pull/122/commits/4acd965cb3641598e4c02845d31dea295e42d705, leaving the current approach alive.

We cherry pick only the addition of the user meta from [here](https://github.com/Automattic/newspack-network/pull/125/commits/bad552daa95133ed767521b6c15616525270b521)

And then we merge and release.

Then we'll have the current implementation preserved, and we'll start to propagate the new events (but not really use them yet)

Then we have a full release cycle to run all the backfillers without any rush.

Then in the next release we merge the new functionalities that rely on these new events.

wdyt?

adekbadek:
That seems pretty complicated. Alternatively, can a site trigger an update to `newspack_network_membership_plans` option if it's empty? This would ensure that it's always there. Even with the backfill, the option might be deleted later by accident.

leogermani:
Closing in favor of https://github.com/Automattic/newspack-network/pull/137

PR #135, opened 2024-09-09
Handle network subscriptions using local data

leogermani:
Redoing https://github.com/Automattic/newspack-network/pull/125 in the new epic

See https://github.com/Automattic/newspack-network/pull/138 for testing instructions

### DIFF
--- a/includes/woocommerce-memberships/class-subscriptions-integration.php
+++ b/includes/woocommerce-memberships/class-subscriptions-integration.php
@@ -61,19 +61,32 @@ class Subscriptions_Integration {
 		$user_network_subs = self::get_user_network_active_subscriptions( $user_id );
 
 		$network_active_subscription = self::user_has_active_subscription_in_network( $user_id, $plan_network_id );
-		if ( $network_active_subscription ) {
-			$user_membership->add_note(
-				sprintf(
-					/* translators: %1$s is the site URL where the user has an active subscription. %2$d is that subscription ID. */
-					__( 'Membership is not cancelled, because there is an active subscription in %1$s linked to a membership plan with the same network id. (Subscription #%2$d)', 'newspack-plugin' ),
-					$network_active_subscription['site'],
-					$network_active_subscription['subscription']['id']
-				)
-			);
+
+		if ( ! $network_active_subscription ) {
+			return $cancel_or_expire;
+		}
+
+		$prevent_cancellation_note_meta_key = 'prevent_cancellation_note_added_timestamp';
+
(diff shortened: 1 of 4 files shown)


tests/unit-tests/test-subscription-integration.php
```diff
@@ -0,0 +1,432 @@
...
+		);
+	}
+
+	/**
+	 * Test get_network_membership_plans
+	 */
+	public function test_get_network_membership_plans() {
+		$method = self::get_method( 'get_network_membership_plans' );
+
+		$result = $method->invoke( null, 'net-1' );
```
adekbadek: It'd be more readable if the variable names were more descriptive, e.g. `$network_membership_plans` instead of `$result`, and `$get_network_membership_plans_method` instead of `$method`.

tests/unit-tests/test-subscription-integration.php
```diff
@@ -0,0 +1,432 @@
...
+	 *
+	 * @return array
+	 */
+	public function subscriptions_includes_plan_data() {
+		return [
+			[
+				'http://site1',
+				'user_with_sub_on_1',
+				'net-1',
+				500,
```
adekbadek: It'd be more readable if these elements were described in a comment, e.g.:

```suggestion
				500, // Expected ID of the subscription.
```

includes/woocommerce-memberships/class-subscriptions-integration.php
```diff
@@ -0,0 +1,194 @@
...
+			return $cancel_or_expire;
+		}
+
+		$user_id = $user->ID;
+
+		$user_network_subs = self::get_user_network_active_subscriptions( $user_id );
+
+		$network_active_subscription = self::user_has_active_subscription_in_network( $user_id, $plan_network_id );
+		if ( $network_active_subscription ) {
+			$user_membership->add_note(
```
adekbadek: In my testing, this note was added five times. Suggestion: store the timestamp of adding a note in post meta of the membership, and only add if not added in the last e.g. 5 minutes.

adekbadek (approved):
Works as described, but the membership note is repeated and tests could use some readability.

includes/woocommerce-memberships/class-subscriptions-integration.php
leogermani: https://github.com/Automattic/newspack-network/pull/135/commits/203ea707e80d33deef50fa771a387932b723646c

tests/unit-tests/test-subscription-integration.php
leogermani: https://github.com/Automattic/newspack-network/pull/135/commits/baf897d2e7e50d5e57dcfc81714c26fb57b9f868

PR #146, opened 2024-11-06
Fix Submenu Highlights 

ronchambers:
### Problem:

The Network plugin has 3 custom post types: Nodes, Subscriptions, and Orders.  For each of these 3 CPTs, the submenu was not being highlighted when on the "add new" screens.  Here is an example from "add new node":

image

### Solution for Nodes CPT:

Added `add_filter( 'submenu_file', [ __CLASS__, 'submenu_file' ] );`

image

### Solution for Subscriptions and Orders:

If I understand correctly the subscriptions and orders CPTs shouldn't allow "add new" anyway since they are imported/synced into the database.  So just remove ability to "add new".

```
'capabilities' => [
    'create_posts' => false, // Removes "add new".
],
```

image

### DIFF
--- a/includes/hub/database/class-orders.php
+++ b/includes/hub/database/class-orders.php
@@ -80,6 +80,9 @@ class Orders {
 			'show_in_menu'     => Network_Admin::PAGE_SLUG,
 			'can_export'       => false,
 			'capability_type'  => 'post',
+			'capabilities'     => [
+				'create_posts' => false, // Removes "add new".
+			],
 			'show_in_rest'     => false,
 			'delete_with_user' => false,
 		);
--- a/includes/hub/database/class-orders.php
+++ b/includes/hub/database/class-orders.php
@@ -81,7 +81,7 @@ class Orders {
 			'can_export'       => false,
 			'capability_type'  => 'post',
 			'capabilities'     => [
-				'create_posts' => false, // Removes "add new".
+				'create_posts' => 'not_a_real_capability', // Set to a fake capability to remove "add new" button.
 			],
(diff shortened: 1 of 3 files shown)


ronchambers:
@leogermani - can you review this?  I hope my assumptions about the Subscriptions and Orders are correct.  I've never used:

```
'capabilities'     => [
    'create_posts' => false, // Removes "add new".
],
```

but it seems to work 😃

includes/hub/database/class-orders.php
```diff
@@ -80,6 +80,9 @@ public static function register_post_type() {
 			'show_in_menu'     => Network_Admin::PAGE_SLUG,
 			'can_export'       => false,
 			'capability_type'  => 'post',
+			'capabilities'     => [
+				'create_posts' => false, // Removes "add new".
```
leogermani: NIT, but even though this end up working, it's not completely correct... The idea here is to inform a string with which capability will be checked to see if a user can create a post in this post type. By default it will check `create_posts` , we can change the suffix in `capabaility_type` for all caps but we are not.

What we usually do when we want to block something for everybody is to return a capability that noboby will ever have. Like "not_allow" or something.

ronchambers: @leogermani Thanks!  I updated it to this:

```
'capabilities'     => [
    'create_posts' => 'not_a_real_capability', // Set to a fake capability to remove "add new" button.
],
```

I tend to be a bit wordy with my variable names and stings, but I can switch to just `not_allow` if you prefer.

leogermani (approved):

PR #155, opened 2024-11-28
feat: content distribution class

miguelpeixe:
1208510338230630-as-1208621948575042/f

Introduces the main class for the Network's content distribution with basic functionality for a post to be distributed across the network according to a specified configuration.

It uses Newspack Network implementation on top of data events for that, so it creates 2 new data events:

## `network_post_updated`

Fires once a distributed post updates, with the following payload:

```json
{
	"site_url": "https://node1.com",
	"post_id": 1,
	"config": {
		"enabled": true,
		"site_urls": [ "https://hub.com", "https://node2.com" ],
		"network_post_id": "1234567890abcdef1234567890abcdef"
	},
	"post_data": {
		"title": "Hello World",
		"date": "2021-01-01 00:00:00",
		"slug": "hello-world",
		"post_type": "post",
		"raw_content": "Content",
		"content": "<p>Content</p>",
		"excerpt": "Excerpt"
	}
}
```

`$post_payload['config']` contains the required information for distribution:
 - `enabled`: Whether distribution is enabled
 - `site_urls`: Sites to distribute the post to
 - `network_post_id`: An md5 composed of site URL and post ID to be unique across the network

The network post ID is used to match an existing linked post to perform updates:

https://github.com/Automattic/newspack-network/blob/2d3f9ca6c2dcfa6ecfcbe06db1896e0fa1c6816c/includes/class-content-distribution.php#L278-L304

## `network_linked_post_inserted`

Fires once a linked post is updated, with the following payload:

```json
{
	"origin": {
		"site_url": "https://node1.com",
		"post_id": 1
	},
	"destination": {
		"site_url": "https://node2.com",
		"post_id": 123,
		"is_unlinked": false
	}
}
```

This event is not yet used but should allow the origin site (or hub) to be aware of its linked posts across the network. This can be for UI or system purposes.

### Testing

1. Make sure you have a working Network setup with at least 1 hub and 2 nodes (let's call them hub, node1, node2)
2. On node1, create a new post with various blocks (paragraph, image, quote, button, HPP...) and grab the post ID
3. Open wp shell and configure the post distribution to hub and node2 by running with the following:

```php
$site_urls = [ 'http://hub.local', 'http://node2.local' ]; // Change to match your setup urls.
Newspack_Network\Content_Distribution::set_post_distribution( $post_id, $site_urls );
```

4. The data event is attached to `wp_after_insert_post`, so you can either hit "Save" on the post or trigger it manually via shell with:

```php
Newspack_Network\Content_Distribution::distribute_post( $post_id );
```

5. Once all the events are dealt with, you should see the post with the correct title, date, and content in both hub and node2. Author and taxonomies are not covered in this PR (see 1208510338230630-as-1208621948575089/f and 1208510338230630-as-1208621948575091/f)
6. Let's unlink the post in node2 by running in node2's shell:

```php
$linked_post_id = 0; // Change to node2's linked post ID
Newspack_Network\Content_Distribution::set_post_unlinked( $linked_post_id );
```

7. In node1, make changes to the post and save
8. Once the events are done, confirm the post has been updated in the hub but remained unchanged in node2
9. Make changes to the post in node2 and save
10. Relink the post in node2:

```php
$linked_post_id = 0; // Change to node2's linked post ID
Newspack_Network\Content_Distribution::set_post_unlinked( $linked_post_id, false );
```

11. Confirm the post in node2 now reflects the most recent content from the distributed post

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -22,9 +22,9 @@ class Content_Distribution {
 	const DISTRIBUTED_POST_META = 'newspack_network_distributed';
 
 	/**
-	 * Post meta key for the linked post containing the distributed post hash.
+	 * Post meta key for the linked post containing the distributed post id.
 	 */
-	const POST_HASH_META = 'newspack_network_post_hash';
+	const NETWORK_POST_ID_META = 'newspack_network_post_id';
 
 	/**
 	 * Post meta key for the linked post containing the distributed post full payload.
@@ -147,9 +147,9 @@ class Content_Distribution {
 		if ( ! is_array( $config ) ) {
 			$config = [];
 		}
-		// Set post hash to link the post across the network.
-		if ( empty( $config['post_hash'] ) ) {
-			$config['post_hash'] = wp_generate_password( 32, false );
+		// Set post network ID.
+		if ( empty( $config['network_post_id'] ) ) {
+			$config['network_post_id'] = md5( $post_id . get_bloginfo( 'url' ) );
 		}
 		$config['enabled']   = empty( $site_urls ) ? false : true;
 		$config['site_urls'] = $site_urls;
@@ -217,9 +217,9 @@ class Content_Distribution {
 		$config = wp_parse_args(
 			$config,
 			[
-				'enabled'   => false,
-				'site_urls' => [],
-				'post_hash' => '',
+				'enabled'         => false,
+				'site_urls'       => [],
+				'network_post_id' => '',
 			]
 		);
 		return $config;
@@ -278,12 +278,12 @@ class Content_Distribution {
 	/**
 	 * Get a linked post given the distributed post hash.
 	 *
-	 * @param string $post_type The post type.
-	 * @param string $post_hash The distributed post hash.
+	 * @param string $post_type       The post type.
+	 * @param string $network_post_id The network post ID for the distributed post.
 	 *
 	 * @return WP_Post|null The linked post or null if not found.
 	 */
-	protected static function get_linked_post( $post_type, $post_hash ) {
+	protected static function get_linked_post( $post_type, $network_post_id ) {
 		$posts = get_posts(
 			[
 				'post_type'      => $post_type,
@@ -291,8 +291,8 @@ class Content_Distribution {
 				'posts_per_page' => 1,
 				'meta_query'     => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
 					[
-						'key'   => self::POST_HASH_META,
-						'value' => $post_hash,
+						'key'   => self::NETWORK_POST_ID_META,
+						'value' => $network_post_id,
 					],
 				],
 			]
@@ -354,7 +354,7 @@ class Content_Distribution {
 		$post_data = $post_payload['post_data'];
 
 		// Post payload is not for a distributed post.
-		if ( ! $config['enabled'] || empty( $config['post_hash'] ) || empty( $config['site_urls'] ) ) {
+		if ( ! $config['enabled'] || empty( $config['network_post_id'] ) || empty( $config['site_urls'] ) ) {
 			return new WP_Error( 'invalid_post_payload', __( 'Invalid post payload.', 'newspack-network' ) );
 		}
 
@@ -364,11 +364,11 @@ class Content_Distribution {
 			return new WP_Error( 'invalid_site', __( 'Post is not configured to be distributed to this site.', 'newspack-network' ) );
 		}
 
-		$post_hash = $config['post_hash'];
+		$network_post_id = $config['network_post_id'];
 		$post_type = $post_data['post_type'];
 
 		// Get the existing linked post.
-		$linked_post = self::get_linked_post( $post_type, $post_hash );
+		$linked_post = self::get_linked_post( $post_type, $network_post_id );
 
 		$postarr = [
 			'ID'           => $linked_post ? $linked_post->ID : 0,
@@ -408,7 +408,7 @@ class Content_Distribution {
 		}
 
 		update_post_meta( $post_id, self::POST_PAYLOAD_META, $post_payload );
-		update_post_meta( $post_id, self::POST_HASH_META, $post_hash );
+		update_post_meta( $post_id, self::NETWORK_POST_ID_META, $network_post_id );
 
 		/**
 		 * Fires after a linked post is inserted.
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -385,7 +385,7 @@ class Content_Distribution {
 
 		// New post, set post status.
 		if ( ! $linked_post ) {
-			$post_data['post_status'] = 'draft';
+			$postarr['post_status'] = 'draft';
 		}
 
 		// Insert the post if it doesn't exist or if it's linked.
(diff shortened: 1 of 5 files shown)


includes/class-content-distribution.php
```diff
@@ -0,0 +1,422 @@
...
+	const POST_UNLINKED_META = 'newspack_network_post_unlinked';
+
+	/**
+	 * Initialize this class and register hooks
+	 *
+	 * @return void
+	 */
+	public static function init() {
+		add_action( 'init', [ __CLASS__, 'register_listeners' ] );
+		add_filter( 'newspack_webhooks_request_priority', [ __CLASS__, 'webhooks_request_priority' ], 10, 2 );
```
miguelpeixe, line 46: This requires https://github.com/Automattic/newspack-plugin/pull/3587, which is currently up for review.

includes/class-content-distribution.php
```diff
@@ -0,0 +1,422 @@
...
+	 * @return void|WP_Error Void on success, WP_Error on failure.
+	 */
+	public static function set_post_unlinked( $post_id, $unlinked = true ) {
+		if ( ! $post_id || ! get_post( $post_id ) ) {
+			return new WP_Error( 'invalid_post', __( 'Invalid post.', 'newspack-network' ) );
+		}
+		update_post_meta( $post_id, self::POST_UNLINKED_META, (bool) $unlinked );
+
+		// If the post is being re-linked, update content.
+		if ( ! $unlinked ) {
```
leogermani, line 324: double negative variables are usually pretty confusing to read. "If Unlinked = false it means that yes, it's linked".. Maybe consider having a LINKED state, with straight forward boolean value that will not be confusing

includes/class-content-distribution.php
```diff
@@ -0,0 +1,422 @@
...
+	/**
+	 * Manually trigger post distribution.
+	 *
+	 * @param WP_Post|int $post The post object or ID.
+	 */
+	public static function distribute_post( $post ) {
+		if ( ! self::is_post_distributed( $post ) ) {
+			return new WP_Error( 'post_not_distributed', __( 'The post is not distributed across the network.', 'newspack-network' ) );
+		}
+		Data_Events::dispatch( 'network_post_updated', self::get_post_payload( $post ) );
```
leogermani: Maybe get the payload from `handle_post_update` so you have one single source of truth about the data payload. (If anything changes in that method, you might end up with discrepancies between the two ways this event is dispatched

includes/class-content-distribution.php
```diff
@@ -0,0 +1,422 @@
...
+
+	/**
+	 * The post meta key for the distributed post configuration.
+	 */
+	const DISTRIBUTED_POST_META = 'newspack_network_distributed';
+
+	/**
+	 * Post meta key for the linked post containing the distributed post hash.
+	 */
+	const POST_HASH_META = 'newspack_network_post_hash';
```
leogermani: When I read post-hash I was expecting something like a hash of site-url plus post ID, so we can verify that it's the same post on both ends. Since you're only doing an ID, maybe it can be called ID? Network id? Also, you can use `uniqid()`

includes/class-content-distribution.php
miguelpeixe: I explored alternatives and I'm not sure it improves readability nor maitainability. Ultimately, the post is always _linked_ and the `insert_linked_post()` method is still called when "unlinked".

Having `unlinked` as the post meta also makes it easier to manage the different state, as we'll replace the content once the flag is removed. Having it the other way around adds some complexity that I don't think justifies the change.

Perhaps we can find a different name for `unlinked` that's not a negative so we can preserve the logic?

includes/class-content-distribution.php
miguelpeixe: Good call! Updated in b0c0a0d9f24182e108170600eaa773495d33d319. In this case, we check on `$data` because the dispatch call does not bail on empty data, like data event listeners.

includes/class-content-distribution.php
miguelpeixe: I like that. Changed to `network_post_id` using an md5 of site URL and post ID in 2d3f9ca6c2dcfa6ecfcbe06db1896e0fa1c6816c

includes/class-content-distribution.php
leogermani: > Perhaps we can find a different name for unlinked that's not a negative so we can preserve the logic?

Yeah, that would be nice, but I can't think of anything good.

Not the end of the world, if you've explored alternatives and think this is more readable let's go with it

miguelpeixe: How about `standalone`?

leogermani: yeah... meh 

What can help as well is if we had helper methods like `set_post_as_linked` ... `is_post_linked`, that would be shortcuts for the `...unlinked( ..., false)` methods

leogermani (approved):
Everything works as described and the implementation is great :) 

post meta is also not coming, but I guess that's still expected.

The only comment I have is that this class is geting big and multi-purpose already. For example we have the registering and handling of the data event, we have the methods to distribute a post, and we also have the methods to ingest a distributed post... I believe this would look (and grow) much better if we had a few smaller classes...

Maybe I would even create a `Distributed_Post` class that you could instantiate and call methods like `get_processed_content()`, `unlink()`, `get_config()`, etc...

miguelpeixe:
Thanks for the review, @leogermani!

> The only comment I have is that this class is geting big and multi-purpose already.

I agree, if you look at the next PRs it's about to get bigger and I'd like to split it up. Once the other PRs are merged, I'll prep a PR solely for organization purposes, with no functional changes, for easier review.

> Maybe I would even create a Distributed_Post class that you could instantiate and call methods like get_processed_content(), unlink(), get_config(), etc...

Yes, some OOP might be good here. I'll see how it looks.

naxoc:
I'm a bit late to the review party, but I have a few ideas:

* The `set_post_distribution()` function naming: I wonder if it would be easier to understand if it was called `set_post_linked()` to match the `set_post_unlinked()` naming? 
* In `set_post_distribution()` – should we validate that the site_urls are valid networked sites?

miguelpeixe:
Thanks for the comments, @naxoc!

> The set_post_distribution() function naming: I wonder if it would be easier to understand if it was called set_post_linked() to match the set_post_unlinked() naming?

I think that would be the case if the method received the same boolean param to toggle link/unlinked. With `set_post_distribution()` we are configuring how the post distribution works.

From Leo's suggestion, I think it'll be better once we move these methods to a `Distributed_Post` with a `set_config()` method. WDYT?

> In set_post_distribution() – should we validate that the site_urls are valid networked sites?

Good call! I know the hub is aware of all the nodes, but I'm not sure it's on node to nodes. If that check requires an API fetch, I'd rather not do it. I'll check!

naxoc:
> From Leo's suggestion, I think it'll be better once we move these methods to a Distributed_Post with a set_config() method. WDYT?

I like that 👍 

I wrote a validation thingy in this PR: https://github.com/Automattic/newspack-network/pull/159/files (line 70-ish) – and then thought that we should maybe centralize it. I can also just do that in the PR if you are ok with that (note that the PR is work in progress).

miguelpeixe:
That's a great idea! I'm working on a refactor for the OOP approach in #160. I can add support for it there so we do as you suggest, a single point of true for this check.

PR #157, opened 2024-12-04
feat(content-distribution): handle post thumbnail

miguelpeixe:
Add support for syncing the post thumbnail when distributing content.

### Testing

1. Make sure you have a network setup with at least 1 hub and 1 node
2. On the hub, create a new post with a featured image, grab the post ID, and use the WP shell to distribute to the node:

```php
$site_urls = [ 'http://node1.local' ]; // Change to match your setup urls.
Newspack_Network\Content_Distribution::set_post_distribution( $post_id, $site_urls );
Newspack_Network\Content_Distribution::distribute_post( $post_id );
```

3. Visit the media library and confirm the post thumbnail was created
4. Confirm the post is added as a draft and the thumbnail is assigned
5. In the original post, change the featured image and save
6. Visit the media library in node1 and confirm that the previous featured image has been deleted and the new one was created
7. Confirm the new featured image was also updated in the post
8. In the original post, delete the featured image and save
9. Back in node1, confirm in the media library that the featured image has been deleted
10. Confirm in the linked post that no featured image is assigned

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -22,9 +22,9 @@ class Content_Distribution {
 	const DISTRIBUTED_POST_META = 'newspack_network_distributed';
 
 	/**
-	 * Post meta key for the linked post containing the distributed post hash.
+	 * Post meta key for the linked post containing the distributed post id.
 	 */
-	const POST_HASH_META = 'newspack_network_post_hash';
+	const NETWORK_POST_ID_META = 'newspack_network_post_id';
 
 	/**
 	 * Post meta key for the linked post containing the distributed post full payload.
@@ -147,9 +147,9 @@ class Content_Distribution {
 		if ( ! is_array( $config ) ) {
 			$config = [];
 		}
-		// Set post hash to link the post across the network.
-		if ( empty( $config['post_hash'] ) ) {
-			$config['post_hash'] = wp_generate_password( 32, false );
+		// Set post network ID.
+		if ( empty( $config['network_post_id'] ) ) {
+			$config['network_post_id'] = md5( $post_id . get_bloginfo( 'url' ) );
 		}
 		$config['enabled']   = empty( $site_urls ) ? false : true;
 		$config['site_urls'] = $site_urls;
@@ -217,9 +217,9 @@ class Content_Distribution {
 		$config = wp_parse_args(
 			$config,
 			[
-				'enabled'   => false,
-				'site_urls' => [],
-				'post_hash' => '',
+				'enabled'         => false,
+				'site_urls'       => [],
+				'network_post_id' => '',
 			]
 		);
 		return $config;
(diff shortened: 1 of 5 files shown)


includes/class-content-distribution.php
```diff
@@ -338,6 +339,44 @@ protected static function is_post_unlinked( $post_id ) {
...
+			$post_payload &&
+			$post_payload['post_data']['thumbnail_url'] === $thumbnail_url
+		) {
+			return;
+		}
+
+		// Delete existing thumbnail attachment.
+		if ( $current_thumbnail_id ) {
+			wp_delete_attachment( $current_thumbnail_id, true );
+		}
```
miguelpeixe: When syncing the post thumbnail, whether the distributed post is updating the featured image or deleting it, the destination site will permanently delete the orphan image.

I'm not sure if this is too aggressive, but I also don't like the idea of lingering orphan images.

The current approach can be problematic if:

- The site starts using the synced attachment in other content
- The site unlinked the post, changed the featured image to something local, and re-linked the post - deleting their local attachment

We can prevent the last one above by adding a meta to the synced attachment so only those get deleted.

naxoc: It's probably a pretty expensive query, but do you think it's worth checking if other posts are using the image before deleting it?

There is also the use case that the image is used in the post body.

includes/class-content-distribution.php
```diff
@@ -339,6 +339,44 @@ protected static function is_post_unlinked( $post_id ) {
...
+
+		if ( ! function_exists( 'media_sideload_image' ) ) {
+			require_once ABSPATH . 'wp-admin/includes/media.php';
+			require_once ABSPATH . 'wp-admin/includes/file.php';
+			require_once ABSPATH . 'wp-admin/includes/image.php';
+		}
+
+		$attachment_id = media_sideload_image( $thumbnail_url, $post_id, '', 'id' );
+		if ( is_wp_error( $attachment_id ) ) {
+			return;
```
naxoc, line 375: A silent error here is probably fine, so this is a question more than a request – but do we have a way to log stuff like this?

naxoc (commented):
This looks good and tests well. (The unit-test is broken, but that is from the updated-date-PR I think).

I added some comments about the image deletion though. I think there are use cases we can get in trouble with.

includes/class-content-distribution.php
miguelpeixe: I've decided to not delete the attachment for now. Instead, we set a post meta to not lose track of the attachments that came via distribution and we can decide later how to deal with them. WDYT? e5d14189d3ba9e60f0bbde88bebe892d92803f0d

I think a post content query is not viable, even if executed via cron. It's very expensive...

naxoc: I think that makes sense. I like the idea with the postmeta. Better have a bit too much media than media missing.

includes/class-content-distribution.php
miguelpeixe: Good call! We have a Debugger class, added a log in 23b93e649d39e0091f7b28eaabd1b6b7f6ab89c5

miguelpeixe:
Thank you for the review, @naxoc!

I've addressed the comments and fixed the tests in 9c7212b648285930ec9f3ea787bceb2a653f6482

naxoc (approved):
Good to go!
ok

PR #159, opened 2024-12-09
feat(content-distribution): add CLI command for distribute post

naxoc:
This adds a CLI command to distribute a post.

I added support for passing 'all' or a list of urls to the `--sites` param, but not site_ids. The original task was ` —sites(all|site_ids|site_urls)`. Is the site ids thing necessary? I feel like it might introduce more error than convenience. I'm happy to add it if we have a use case that makes sense.

I didn't add tests because it doesn't make much sense to test CLI commands that just calls code that has tests.

This one relies on #164 being merged for sync to work.

# How to test
```bash
# Create a post on the hub:
wp post create --post_content="Oh, wow" --post_status=publish --post_title="I'll go places from $(wp option get blogname)"
# Grab the ID and call this:
wp newspack network distribute post <post-id> --sites=https://node1.test
# Check that the post shows up.

# Now do it again with one more site to test the adding of sites:
wp newspack network distribute post <post-id> --sites=https://node2.test

# Create another post and test with 'all'
wp newspack network distribute post <post-id> --sites=all

# Now find a page ID and run the below. It should throw an error
wp newspack network distribute post <page-id> --sites=https://google.com
```


includes/cli/class-distribution.php
```diff
@@ -0,0 +1,107 @@
...
+				fn( $site ) => untrailingslashit( trim( $site ) ),
+				explode( ',', $assoc_args['sites'] )
+			);
+			$urls_not_in_network = array_filter(
+				$sites,
+				fn( $site ) => ! Network::is_networked_url( $site )
+			);
+			if ( ! empty( $urls_not_in_network ) ) {
+				WP_CLI::error( sprintf( 'Non-networked URLs were passed in --sites: %s', implode( ', ', $urls_not_in_network ) ) );
+			}
```
miguelpeixe: This seems like a useful logic to be in `Outgoing_Post->set_config()` instead of here. WDYT?

includes/cli/class-distribution.php
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		try {
+			$outgoing_post = Content_Distribution::get_distributed_post( $post_id ) ?? new Outgoing_Post( $post_id );
+			$outgoing_post->set_config( $sites );
```
miguelpeixe: I wonder if it'd be expected behavior for a command to overwrite any previously configured sites. I can see a case where a post is being distributed to a site and the admin would like to add a new one when running the command.

includes/content-distribution/class-outgoing-post.php
```diff
@@ -40,6 +40,11 @@ public function __construct( $post ) {
 			throw new \InvalidArgumentException( esc_html( __( 'Invalid post.', 'newspack-network' ) ) );
 		}
 
+		if ( ! in_array( $post->post_type, Content_Distribution::get_distributed_post_types() ) ) {
+			/* translators: unsupported post type for content distribution */
+			throw new \InvalidArgumentException( esc_html( sprintf( __( 'Post type %s is not supported as a distributed outgoing post.', 'newspack-network' ), $post->post_type ) ) );
+		}
+
```
miguelpeixe, line 48: 💯

miguelpeixe (changes requested):
Tests well and the code looks good!

I have a couple of questions inline

includes/cli/class-distribution.php
naxoc: That makes sense. I'll move it!

includes/cli/class-distribution.php
naxoc: Yeah, I was not sure either. It might be better if I only add the sites missing in the config?

miguelpeixe: It might, but that would warrant a way to remove sites from the config.

Maybe we can leave `--sites` as is and add a `--add-sites` to append.

includes/cli/class-distribution.php
naxoc: I'm now debug logging it if a url was passed that is not networked.

includes/cli/class-distribution.php
naxoc: I made it so we can add sites with `--sites`. We can't remove ones though.

naxoc:
OK, @miguelpeixe – I had to update the test a bit with some mockery, but I think this is ready for re-review. 

Note that the `set_config()` now returns the config – I use it to write a message to the CLI. Let me know if that makes no sense.

includes/content-distribution/class-outgoing-post.php
```diff
@@ -68,8 +92,18 @@ public function set_config( $site_urls = [] ) {
...
 		}
-		$config['site_urls'] = $site_urls;
+
+		if ( empty( $config['site_urls'] ) ) {
+			$config['site_urls'] = [];
+		}
+
+		// If there are urls not already in the config, add them. Note that we don't support
+		// removing urls from the config.
+		$config['site_urls'] = array_unique( array_merge( $config['site_urls'], $site_urls ) );
```
miguelpeixe, line 102: Can we unit test this new thing?

miguelpeixe (changes requested):
That's perfect! Thank you for all the revisions 🎉 

One nitpicky thing below

includes/content-distribution/class-outgoing-post.php
naxoc: Yep! I added a test 👍

miguelpeixe (approved):

PR #160, opened 2024-12-09
refactor: OOP for distributed and linked posts

miguelpeixe:
Refactors the Content Distribution class to use instances of the new `Outgoing_Post` and `Incoming_Post` classes.

The `Outgoing_Post` is initialized with a post object or ID:

```php
$post_id = 1; // ID for a distributed post.
$config = $outgoing_post->get_config();
// = [
//     "site_urls" => [
//       "http://node1.local",
//     ],
//     "network_post_id" => "0ac636e2f9dfc3ddd9d1934bfa97b252",
//  ]

$post_id = 2; // ID for a non-distributed post.
$config = $outgoing_post->get_config();
// = [
//     "site_urls" => [],
//     "network_post_id" => "",
//  ]
```

The `Incoming_Post` is initialized with a post payload (`$outgoing_post->get_payload()`) and queries the existing linked post on instantiation:

```php
$incoming_post = new Newspack_Network\Content_Distribution\Incoming_Post( $payload );

// Insert the post with the object payload.
$post_id = $incoming_post->insert();
```

The methods and unit tests had most of their logic preserved with some adjustments to fit the new approach. The `Content_Distribution` class is still responsible for implementing the hooks that make the distribution possible, connecting the new objects.

This PR also implements the support for syncing taxonomies (merged from #158). Testing this new feature is a good way to ensure that this refactor didn't introduce any regression to the core elements of the project.

### Testing

1. Make sure you have a network setup with at least 1 hub and 1 node
2. On the hub, create a new post with a featured image, categories, and tags
3. Grab the post ID, and use the WP shell to distribute to the node:

```php
$outgoing_post->set_config( [ 'http://node1.local' ] ); // Change to match your node URL.
```

4. Once the events are handled, visit the node's drafts and confirm the post is listed with the correct taxonomy and featured image
5. Back in the hub, change the categories and terms and save
6. In the node, confirm the taxonomy changes were applied


includes/content-distribution/class-distributed-post.php
```diff
@@ -0,0 +1,203 @@
...
+	 * The post meta key for the distributed post configuration.
+	 */
+	const DISTRIBUTED_POST_META = 'newspack_network_distributed';
+
+	/**
+	 * The post ID.
+	 *
+	 * @var int
+	 */
+	public $ID = 0;
```
naxoc: Is this necessary? we have $this->post->ID

includes/content-distribution/class-distributed-post.php
```diff
@@ -0,0 +1,203 @@
...
+		}
+		$config = get_post_meta( $this->ID, self::DISTRIBUTED_POST_META, true );
+		if ( ! is_array( $config ) ) {
+			$config = [];
+		}
+		// Set post network ID.
+		if ( empty( $config['network_post_id'] ) ) {
+			$config['network_post_id'] = md5( $this->ID . get_bloginfo( 'url' ) );
+		}
+		$config['enabled']   = empty( $site_urls ) ? false : true;
```
naxoc: I struggle understanding`enabled` a bit. Why would we have a disabled distributed post? Is this the same as linked?

includes/content-distribution/class-distributed-post.php
```diff
@@ -0,0 +1,203 @@
...
+		$config['site_urls'] = $site_urls;
+		update_post_meta( $this->ID, self::DISTRIBUTED_POST_META, $config );
+	}
+
+	/**
+	 * Whether the post is distributed. Optionally provide a $site_url to check if
+	 * the post is distributed to that site.
+	 *
+	 * @param int|null $site_url Optional site ID.
```
naxoc: Should that not be string? And url? Not ID?

includes/content-distribution/class-distributed-post.php
```diff
@@ -0,0 +1,203 @@
...
+	 *
+	 * @return bool
+	 */
+	public function is_distributed( $site_url = null ) {
+		if ( ! $this->ID ) {
```
naxoc: This would never happen I think? It's set in the constructor. This goes for all the checks.

includes/content-distribution/class-linked-post.php
```diff
@@ -0,0 +1,395 @@
...
+			! is_array( $payload ) ||
+			empty( $payload['post_id'] ) ||
+			empty( $payload['config'] ) ||
+			empty( $payload['post_data'] )
+		) {
+			return new WP_Error( 'invalid_post', __( 'Invalid post payload.', 'newspack-network' ) );
+		}
+
+		$config    = $payload['config'];
+		$post_data = $payload['post_data'];
```
naxoc: This var is unused

includes/content-distribution/class-linked-post.php
```diff
@@ -0,0 +1,395 @@
...
+
+		if ( is_wp_error( $error ) ) {
+			throw new \InvalidArgumentException( esc_html( $error->get_error_message() ) );
+		}
+
+		$this->payload         = $payload;
+		$this->network_post_id = $payload['config']['network_post_id'];
+
+		$post = $this->query_linked_post();
+		if ( $post ) {
```
naxoc: Would it be better to throw an error here or is there a use case for having no post?

naxoc (commented):
This looks good! Much better separation.

I think I'm still learning, so I added some questions that might be silly about the Linked post. I don't understand what a linked post that doesn't have a post is or should do.

includes/content-distribution/class-linked-post.php
miguelpeixe: The `Linked_Post` can - and will - be instantiated without the post existing. It will exist once `->insert()` is called.

includes/content-distribution/class-linked-post.php
miguelpeixe: Whoops! Inherited from its previous home. Fixed in 086ec9cdea66aa7705c405f0609fc43fef68dff4

includes/content-distribution/class-distributed-post.php
miguelpeixe: Thanks for catching that! Fixed in 2bb94b6ab310c07376683cafcb197d9e685c98fa

naxoc:
I am getting an error when clicking "Add New Post" in `wp-admin/edit.php`

```
PHP Fatal error:  Uncaught Error: Call to a member function is_distributed() on null in /Users/ckj/Dev/_code/re│Stack trace:
│#0 [internal function]: Newspack_Network\Content_Distribution::handle_post_updated(NULL, Object(WP_Post), false, NULL)
│#1 /Users/ckj/Dev/_code/repos/newspack-plugin/includes/data-events/class-data-events.php(192): call_user_func_array(Array, Array)
│#2 /Users/ckj/Dev/sites/hub/wp-includes/class-wp-hook.php(324): Newspack\Data_Events::Newspack\{closure}(140, Object(WP_Post), false, NULL)
│#3 /Users/ckj/Dev/sites/hub/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters(NULL, Array)
│#4 /Users/ckj/Dev/sites/hub/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
│#5 /Users/ckj/Dev/sites/hub/wp-includes/post.php(5781): do_action('wp_after_insert...', 140, Object(WP_Post), false, NULL)
│#6 /Users/ckj/Dev/sites/hub/wp-admin/includes/post.php(772): wp_after_insert_post(Object(WP_Post), false, NULL)
│#7 /Users/ckj/Dev/sites/hub/wp-admin/post-new.php(66): get_default_post_to_edit('post', true)
│#8 /Users/ckj/.composer/vendor/laravel/valet/server.php(110): require('/Users/ckj/Dev/...')
│#9 {main}
│  thrown in /Users/ckj/Dev/_code/repos/newspack-network/includes/class-content-distribution.php on line 69
```

includes/content-distribution/class-distributed-post.php
miguelpeixe: It's a public var, whereas `$post` is protected. To avoid the duplication I changed the strategy to deprecate this var and add a public `get_post()` in f755a01ccc65854929b8402913e149f49b1853e1. WDYT?

includes/content-distribution/class-distributed-post.php
miguelpeixe: You're right! Removed in f755a01ccc65854929b8402913e149f49b1853e1

includes/content-distribution/class-distributed-post.php
naxoc: I like it! 👍

includes/content-distribution/class-distributed-post.php
miguelpeixe: It's mostly a proposal to be able to deactivate the distribution of a given post without removing the configured site list. I can see this being a toggle in the post distribution configuration, but that's not really scoped.

miguelpeixe:
> I am getting an error when clicking "Add New Post" in wp-admin/edit.php

Caused by a last-minute change in the `Content_Distribution::handle_post_updated()` method 🤦 

Fixed in dddda16d9744d92f76617675f7e797a091c67b44

includes/content-distribution/class-distributed-post.php
miguelpeixe: Missed one: 8bafa68023772b7abe6d20f7540e23034f748315

includes/content-distribution/class-distributed-post.php
naxoc: So it means "distribution is enabled"? 

> to be able to deactivate the distribution of a given post without removing the configured site list. 

But here it is based on not having sites in the list. 

Would it be easier to understand if it was a count of the site urls? (and a different name)?

miguelpeixe: Currently it's an automated way to set `enabled` given the site list.

There can be a `deactivate_distribution()` method that just toggles that prop. There's no need for it right now. I can also just remove the prop entirely and we implement once there's a case for it.

miguelpeixe: Updated in 22c6e8fd739a7776a0b0c16b125f8bc9597d0693

miguelpeixe:
As discussed in p1733836852714589-slack-C055SRT7WN4, the classes have been renamed to `Outgoing_Post` and `Incoming_Post` in ef68219ee064b3105eead1b6d361dfb4e91d38f8

naxoc (approved):
I think this looks good now! Thank you for your patience with all my questions, Miguel 🙌

miguelpeixe:
Thank **you** for the thorough review and great questions 🙌

PR #161, opened 2024-12-10
feat: Add network util class

naxoc:
So far the class has a method to get all networked urls and one to check if a url is in the network.

I didn't write tests for this because I'm unsure if that is even possible without a very complex unittest-framework?

To test it – it's probably easiest to just call it directly. 
```bash
# Do this on both the hub and a node or two. The url of the site you are calling it on should not be included.
wp eval "var_dump(Newspack_Network\Utils\Network::get_networked_urls());"

# Should be false:
wp eval "var_dump(Newspack_Network\Utils\Network::is_networked_url('https://google.com'));"
# Should be true:
 wp eval "var_dump(Newspack_Network\Utils\Network::is_networked_url(home_url()));"
# Also try with other urls in your network.
```

**Note** I had to update the PHP version in `phpcs.xml` to allow for me using the splat operator in `get_networked_urls()`.


miguelpeixe (approved):
Nice approach and works without issues!

I'd like @leogermani input on the modern syntax, since we're not using it anywhere in the codebase yet. IMO, if it's 8.1+ compliant we should be fine.

PR #162, opened 2024-12-10
feat: place content distribution behind a constant

miguelpeixe:
Place the feature behind a constant so it can be merged to trunk and we can work outside of epic branches.

As we are still without any UI, the constant will only prevent hooks from allowing distribution to occur. The classes and methods should remain available via code.

### Testing

1. Set `define( 'NEWPACK_NETWORK_CONTENT_DISTRIBUTION', true );` in your `wp-config.php`
2. Make sure you have a network setup with at least 1 hub and 1 node
3. On the hub, create a new post and configure its distribution:

```php
$outgoing_post->set_config( [ 'http://node1.local' ] ); // Change to match your node URL.
```

4. Confirm the node receives the post
5. Change the constant to false in `wp-config.php`
6. Make a change to the original post and save
7. Confirm the changes are not distributed

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -22,6 +22,9 @@ class Content_Distribution {
 	 * @return void
 	 */
 	public static function init() {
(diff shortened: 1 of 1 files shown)


naxoc (approved):
Works! 👍

PR #165, opened 2024-12-11
fix(content-distribution): debug post update and remove deprecated config

miguelpeixe:
Add debugger log to the `network_post_updated` event and fixes an issue with `get_payload_error()` using a deprecated config parameter.

### Testing

Confirm that distribution works and the node logs `Processing network_post_updated [config]`


naxoc (approved):
Does the trick. Thanks!
ok

PR #166, opened 2024-12-11
feat(content-distribution): handle status changes

miguelpeixe:
Handle content distribution status changes and deletions.

When a distributed post gets its status changed to non-publish (`draft`, `trash`, `private`, or `pending`), its linked posts should follow.

When a distributed post gets deleted, a linked post should move to the trash and the unlinked post remain with its status. The distributed post payload meta will be deleted as a way to treat the post as a regular standalone post. The network post ID and unlinked meta also remain so if the original post gets restored from a backup, its relationship in the network is also restored.

### Testing

1. Make sure you have a network setup with at least a hub and 2 nodes (let's call node1 and node2)
2. Create a post in the hub and distribute:

```php
$outgoing_post->set_config( [ 'http://node1.local', 'http://node2.local' ] ); // Change to match your nodes.
```

3. Once it is distributed, publish the post in node1 and node2
4. In node2, unlink the post via shell:

```php
$post_id = 0; // Change to the post ID in node2.
$incoming_post = new Newspack_Network\Content_Distribution\Linked_Post( $post_id );
$incoming_post->set_unlinked();
```

5. In the hub, change the post status to `draft`
6. Confirm it moves to draft in node1 and remains published in node2
7. In the hub, delete the post
8. Confirm it moves to the trash in node1 and remains published in node2
9. In both node1 and node2, confirm the post payload meta is gone:

```php
get_post_meta( $post_id, 'newspack_network_post_payload', true );
```


includes/class-content-distribution.php
```diff
@@ -172,6 +174,24 @@ public static function handle_post_updated( $post ) {
 		self::$queued_post_updates[] = $post->ID;
 	}
 
+	/**
+	 * Post delete listener callback.
+	 *
+	 * @param int $post_id The post ID.
+	 *
+	 * @return array|null The post payload or null if the post is not distributed.
```
naxoc: This will never return array – should it return or can we fix the comment?

naxoc (commented):
This tests well and looks good. I could not get the last step to (check that post payload meta is gone) to work. I still have it around. 

Am I missing something or can you also not get that deleted?

includes/class-content-distribution.php
miguelpeixe: Whoops, that's reminiscent from the previous approach. Thank you for catching that. Updated in 647d24710b2fcf595b29220cc9de093848b2eb83

miguelpeixe:
> I could not get the last step to (check that post payload meta is gone) to work. I still have it around.
> 
> Am I missing something or can you also not get that deleted?

Hmm.. that can bite us in the future. I'll double-check what could be happening here.

miguelpeixe:
That's weird. I don't know what could be getting in the way and unit tests also pass that step.

Can you confirm the distributed post got permanently deleted and not just in the trash? The trash status will also trash a linked post, but not delete the post meta.

naxoc:
> Can you confirm the distributed post got permanently deleted and not just in the trash? The trash status will also trash a linked post, but not delete the post meta.

Woops. I did just trash and not delete. This is going to sound stupid, but how do I delete without trashing it? Deleting it after trashing didn't trigger an event I think?

miguelpeixe:
It should trigger from trash to deletion, as long as you use wp for that and not directly in the db

includes/class-content-distribution.php
```diff
@@ -172,6 +174,24 @@ public static function handle_post_updated( $post ) {
...
+	 *
+	 * @return @void
+	 */
+	public static function handle_post_deleted( $post_id ) {
+		if ( ! class_exists( 'Newspack\Data_Events' ) ) {
+			return;
+		}
+		$post = self::get_distributed_post( $post_id );
+		if ( ! $post ) {
+			return;
```
naxoc, line 190: I don't think I'm able to get the code past here. The post is already deleted?

naxoc (approved):
Works like a charm now. Thanks!

PR #167, opened 2024-12-11
feat(content-distribution): editor plugin for distribution

miguelpeixe:
Implements an editor plugin for distributing a post.

image

My original intention was to strictly leverage `editPost()` and `savePost()` dispatches to update the distribution post meta and let our short-circuit hook handle validation. This did not work well for error handling, as there was no straightforward way to pass the error message to the editor. Because of that, I switched to an API endpoint.

### Selecting connections and distributing

newspack sa ngrok io_wp-admin_post php_post=328 action=edit

newspack sa ngrok io_wp-admin_post php_post=328 action=edit (1)

newspack sa ngrok io_wp-admin_post php_post=328 action=edit (2)

newspack sa ngrok io_wp-admin_post php_post=328 action=edit

newspack sa ngrok io_wp-admin_post php_post=328 action=edit (3)

### Updating connections

newspack sa ngrok io_wp-admin_post php_post=326 action=edit

newspack sa ngrok io_wp-admin_post php_post=326 action=edit (1)

newspack sa ngrok io_wp-admin_post php_post=326 action=edit

newspack sa ngrok io_wp-admin_post php_post=326 action=edit (3)

### Error message

newspack sa ngrok io_wp-admin_post php_post=327 action=edit

### Distributing an unsaved post

To prevent distributing an outdated version of the post, if the post has unsaved changes we'll enforce a post save before distributing:

newspack sa ngrok io_wp-admin_post php_post=352 action=edit

### A lot of nodes

newspack sa ngrok io_wp-admin_post php_post=352 action=edit (1)

## Testing

**This PR also inaugurates javascript assets to the plugin. Before anything, make sure to run `npm ci && npm run build`**

1. Make sure you have a network setup with a hub and 2 nodes and `define( 'NEWPACK_NETWORK_CONTENT_DISTRIBUTION', true );` on all instances config
2. Draft a new post in the hub and click the globe icon as referenced in the first image above
3. Confirm the nodes are available for distribution
4. Make changes to the post and confirm the button changes from "Distribute" to "Save & Distribute"
5. Save the post and confirm the button is back to "Distribute"
6. Select one node and click "Distribute"
7. Confirm you get the pending state as in the image above and the snackbar notice on success
8. Confirm the selected node is now checked and disabled
9. Click "Select all" and confirm the other node gets selected
10. Click Distribute and confirm you get the same successful result
11. Confirm on the nodes that the post is distributed
12. Edit the post in one of the nodes and confirm the editor plugin (globe icon) is not available

### Multiple nodes

1. Replace line 57 of `src/content-distribution/distribute/index.js` with:
```js
	const sites = [];
	for( let i = 0; i < 100; i++ ) {
		sites.push( `https://node${ i + 1}.local` );
	}
```
2. Edit a post and confirm the plugin sidebar renders with a scrollbar

### Error handling

1. Add the following to line 78 of `includes/content-distribution/class-outgoing-post.php`: 
```php
return new WP_Error( 'test_error', 'This is my error message. Try distributing again.' );
```
2. Attempt to distribute a post via the editor and confirm you get the error notice.


src/content-distribution/distribute/style.scss
```diff
@@ -0,0 +1,53 @@
...
+	.components-panel {
+		height: 100%;
+		display: flex;
+		flex-direction: column;
+	}
+	.components-panel__body {
+		border: 0;
+		&.distribute-header {
+			p {
+				color: #757575;
```
thomasguillot: Instead of hardcoding colors, could we use WordPress color variables? https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_colors.scss#L11

src/content-distribution/distribute/index.js
```diff
@@ -0,0 +1,209 @@
...
+				urls: siteSelection,
+			},
+		} ).then( urls => {
+			setDistribution( urls );
+			setSiteSelection( [] );
+			createNotice(
+				'info',
+				sprintf(
+					_n(
+						'Post distributed with one connection.',
```
naxoc: Small language nit: Should this be "Post distributed to"? Distribute with sounds more like shared with

src/content-distribution/distribute/index.js
```diff
@@ -0,0 +1,209 @@
...
+							if ( hasChangedContent || isCleanNewPost ) {
+								savePost().then( distribute );
+							} else {
+								distribute();
+							}
+						} }
+					>
+						{ isDistributing ? __( 'Distributing...', 'newspack-network' ) : (
+							hasChangedContent || isCleanNewPost ?
+							__( 'Save & Distribute', 'newspack-network' ) :
```
naxoc, line 231: I was a little confused about the "Save and Distribute" button. It was unclear to me if that would publish the note or not. I realize there might not be room for "Save draft and distribute" on the button though 🤔

includes/content-distribution/class-api.php
```diff
@@ -0,0 +1,76 @@
...
+	 * @return \WP_REST_Response
+	 */
+	public static function distribute( $request ) {
+		$post_id = $request->get_param( 'post_id' );
+		$urls    = $request->get_param( 'urls' );
+
+		try {
+			$outgoing_post = new Outgoing_Post( $post_id );
+		} catch ( \InvalidArgumentException $e ) {
+			return new WP_Error( 'newspack_network_content_distribution_error', $e->getMessage(), [ 'status' => 400 ] );
```
naxoc, line 64: We should also wrap the errors in rest_ensure_response()

includes/content-distribution/class-editor.php
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Newspack Network Content Distribution Editor.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network\Content_Distribution;
+
+use Newspack_Network\Content_Distribution;
+use Newspack_Network\Content_Distribution\Outgoing_Post;
```
naxoc: This is unused

naxoc (commented):
This is beautiful. Great work!

I only have some UI and language questions and very few nits :) 

Step 12:
> Edit the post in one of the nodes and confirm the editor plugin (globe icon) is not available

I wonder if it is easier to understand if the icon is there, but shows an explanation to why it will not let you sync?

includes/content-distribution/class-editor.php
miguelpeixe: Whoops! If only I used phpstorm 😄 

Updated in 1fafddf4e635cad03de6fb30470d043f3555f589

includes/content-distribution/class-api.php
miguelpeixe: `WP_Error` objects are handled differently by the rest API controller. This function will not process it: https://github.com/WordPress/wordpress-develop/blob/e9dfa8c34c74cb51da93ae4b4e61c1df0378b3d7/src/wp-includes/rest-api.php#L659-L662

src/content-distribution/distribute/style.scss
miguelpeixe: Yes! Updated in aa34cf4ce1a1d20fc124550e49849cf47978da31

miguelpeixe:
I'll defer Camilla's UI questions to @thomasguillot:

- Should the globe icon and sidebar appear for linked posts but with some information regarding why it's disabled? Should we do the same for non-supported post types?
- Should "Post distributed with one connection." be "Post distributed to"? Distribute with sounds more like shared with
- I was a little confused about the "Save and Distribute" button. It was unclear to me if that would publish the note or not. I realize there might not be room for "Save draft and distribute" on the button though 🤔

src/content-distribution/distribute/style.scss
```diff
@@ -0,0 +1,55 @@
...
+	.components-panel__body {
+		border: 0;
+		&.distribute-header {
+			p {
+				color: colors.$gray-700;
+			}
+			.components-base-control__field {
+				margin: 0;
+			}
+			input[type="text"] {
```
thomasguillot: Instead of custom css here, I'd suggest you add `__next40pxDefaultSize` to the `TextControl` (see: https://wordpress.github.io/gutenberg/?path=/docs/components-textcontrol--docs)

miguelpeixe: TIL! Updated in bd46bb52aef86d061549b0b62c8a84104e5da8d1

miguelpeixe:
One other thing I wondered about and would like @thomasguillot's input. When editing a post, it's unclear whether the post is distributed unless you click the globe icon. Should we do something similar to what's proposed for the linked post and add a notice for this:

image

src/content-distribution/distribute/style.scss
```diff
@@ -0,0 +1,55 @@
...
+				padding: 0 16px;
+				height: 40px;
+			}
+		}
+		&.distribute-body {
+			flex: 1 1 100%;
+			overflow-y: auto;
+			padding-top: 0;
+			.components-checkbox-control {
+				--checkbox-input-size: 20px;
```
thomasguillot: No need to tweak the size. I got the size wrong in Figma 🙊

src/content-distribution/distribute/style.scss
```diff
@@ -0,0 +1,55 @@
...
+				height: 44px;
+				display: flex;
+				align-items: center;
+				.components-base-control__field {
+					margin-bottom: 0;
+				}
+				.components-checkbox-control__input-container {
+					margin-right: 1em;
```
thomasguillot: This can be removed. Core's right margin is fine.

includes/content-distribution/class-api.php
naxoc: Can we fix the return type then please? My fancy editor is freaking out ;)

src/content-distribution/distribute/style.scss
miguelpeixe: Ah, ok! I did like the larger checkbox for this, though.

Comparison

| 20px | 16px |
| --- | --- |
| image | image |

includes/content-distribution/class-api.php
miguelpeixe: Ah I see! Absolutely, I'd never beef with phpstorm

miguelpeixe: 35c1b46a094b96ffde5ccf61fbdf46ac4d089505

src/content-distribution/distribute/style.scss
thomasguillot: Hmm yeah but for consistency I prefer to avoid any sort of custom css on all the components

miguelpeixe: Sounds good! Updated in a1a94cafc0997adef87e181488ef094a4eb13bc1

src/content-distribution/distribute/style.scss
miguelpeixe: Updated in a1a94cafc0997adef87e181488ef094a4eb13bc1

src/content-distribution/distribute/style.scss
```diff
@@ -0,0 +1,55 @@
...
+	.components-panel {
+		height: 100%;
+		display: flex;
+		flex-direction: column;
+	}
+	.components-panel__body {
+		border: 0;
+		&.distribute-header {
+			p {
+				color: colors.$gray-700;
```
thomasguillot: Sorry to be a pain. Can we just remove this? I tested with the default gray-900 and I actually prefer it.

miguelpeixe: Sure, I prefer it too! e98d9899e8944c9250f6d48ec75602574034ece1

includes/class-content-distribution.php
```diff
@@ -9,9 +9,12 @@
 
 use Newspack\Data_Events;
 use Newspack_Network\Content_Distribution\CLI;
+use Newspack_Network\Content_Distribution\API;
+use Newspack_Network\Content_Distribution\Editor;
 use Newspack_Network\Content_Distribution\Incoming_Post;
 use Newspack_Network\Content_Distribution\Outgoing_Post;
 use WP_Post;
+use WP_Error;
```
naxoc: This one is unused ;)

src/content-distribution/distribute/style.scss
```diff
@@ -0,0 +1,42 @@
...
+			flex: 1 1 100%;
+			overflow-y: auto;
+			padding-top: 0;
+			.components-checkbox-control {
+				height: 44px;
+				display: flex;
+				align-items: center;
+				.components-base-control__field {
+					margin-bottom: 0;
+				}
```
thomasguillot: Let’s remove all the custom CSS. While it looks better now, I prefer to avoid custom CSS altogether. I must have copied the wrong components from Figma.

thomasguillot: Ignore me

thomasguillot:
I have pushed a tiny change: https://github.com/Automattic/newspack-network/pull/167/commits/6dd77ed27e66d4751b3e143b41adfb5127d148fa

thomasguillot:
Screenshot 2024-12-18 at 15 02 58@2x

@miguelpeixe, what do you think about these suggestions:

- If there's only one node, we should simplify the sidebar by removing the search and "select all" option (and update the description).
- For five or fewer nodes, we should hide the search feature since it's unnecessary.
- Should we truncate long node names?

thomasguillot:
> One other thing I wondered about and would like @thomasguillot's input. When editing a post, it's unclear whether the post is distributed unless you click the globe icon. Should we do something similar to what's proposed for the linked post and add a notice for this

I aimed to keep this as simple as possible, showing only the distribution locations via the sidebar. I understand your point though and agree. I'd just suggest two tweaks:

- Add the globe icon
- Make the notice dismissible

src/content-distribution/distribute/index.js
thomasguillot: You're right, it should be "to" cc @miguelpeixe

thomasguillot:
I noticed that the "Save & Distribute" button is active for unpublished posts. This seems incorrect—shouldn't the button be disabled until after publication?

thomasguillot:
image
image

I don't see these when looking at the distributed node. Am I missing something?

miguelpeixe:
> This seems incorrect—shouldn't the button be disabled until after publication?

Good point, I'm not sure how we want that to work. Technically it's not required for a post to be published to be distributed, but we can enforce that if we want.

If we are to do that, I think it deserves some messaging as to why the button is disabled, otherwise users might get confused.

Wdyt @leogermani  and @naxoc?

miguelpeixe:
> I don't see these when looking at the distributed node. Am I missing something?

That's not covered in the PR

src/content-distribution/distribute/index.js
miguelpeixe: Updated in 8b9bccc6e720c68da37b374bd98318c500b7db41

thomasguillot:
> Should the globe icon and sidebar appear for linked posts but with some information regarding why it's disabled? Should we do the same for non-supported post types?

I think this is https://github.com/Automattic/newspack-network/pull/167#issuecomment-2551712742?

miguelpeixe:
> I think this is https://github.com/Automattic/newspack-network/pull/167#issuecomment-2551712742?

Yes it is :)

thomasguillot:
> I was a little confused about the "Save and Distribute" button. It was unclear to me if that would publish the note or not. I realize there might not be room for "Save draft and distribute" on the button though

It's only when the post has unsaved changes so I think the wording is OK. We should just make sure in the documentation we mention the post will be saved if it has unsaved changes.

miguelpeixe:
> If there's only one node, we should simplify the sidebar by removing the search and "select all" option (and update the description).
> For five or fewer nodes, we should hide the search feature since it's unnecessary.
> Should we truncate long node names?

These changes have been applied in 2e9222b612af267516b2881d3f60779d036dcdac and 74fd79e6dfe48bffce361e6758f1aaf7e4ba0e59

Best way to test it is to replace line 20 of `src/content-distribution/distribute/index.js` with:

```js
const networkSites = [];
const siteAmount = 10;
for ( let i = 0; i < siteAmount; i++ ) {
	networkSites.push( `https://site-${ i }-with-really-very-long-domain-name.com` );
}
```

Then, modify the `siteAmount` value to 1, 2, and 5 to test all the scenarios.

miguelpeixe:
Also, the notice has been added in e10819e4cf5793664fbd8ec35b0710737c553079

leogermani:
>  Technically it's not required for a post to be published to be distributed, but we can enforce that if we want.

We have talked about this before and concluded that we do want this! That removes some complexity on status handling... and also it's how Distributor does it

miguelpeixe:
> We have talked about this before and concluded that we do want this! That removes some complexity on status handling... and also it's how Distributor does it

Sounds good! How about messaging? We just leave the button disabled or do we add something extra?

miguelpeixe:
Added the non-publish post restriction in 0f7f73d625baad3cc941eff283fada06f8243bfc

miguelpeixe:
> > We have talked about this before and concluded that we do want this! That removes some complexity on status handling... and also it's how Distributor does it
> 
> Sounds good! How about messaging? We just leave the button disabled or do we add something extra?

Or maybe we do a "Publish & Distribute" in this case, instead of disabling it 🤔

thomasguillot:
> Or maybe we do a "Publish & Distribute" in this case, instead of disabling it

Too dangerous. A publisher might find useful the "pre-publish checks" sidebar and add tags etc...

thomasguillot:
>Sounds good! How about messaging? We just leave the button disabled or do we add something extra?

I'd change the description to something like: "This [post_type_singularname] has not been published yet. Please publish the [post_type_singularname] before distributing it to any connections."

leogermani:
> I'd change the description to something like: "This [post_type_singularname] has not been published yet. Please publish the [post_type_singularname] before distributing it to any connections."

Ah, I meant to comment on that. "Connections" is a Distributor plugin concept. Not sure if we need/want to use that here... they are "sites in the network"... or just "sites"

miguelpeixe:
"Network site(s)" is a common occurrence. Maybe that?

miguelpeixe:
Here's an example with the new paragraph using the proposed "network site" label:

image

Updated in 5f57b6a4e00cca657bdf3c7a9cadac0775e24d26, let me know if I should tweak the "connection/network site" labeling.

cc @leogermani @thomasguillot

leogermani:
For the yellow notice saying that the post is distributed to other sites, I think it will break the UI if the list has too many sites (not that many actually)... Maybe just display the number of sites? Is it possible to add a "Manage" link that opens the sidebar panel?

miguelpeixe:
> For the yellow notice saying that the post is distributed to other sites, I think it will break the UI if the list has too many sites (not that many actually)... Maybe just display the number of sites?

Good call! Updated in 649cf97a40fc187d96ff61f5b086ca98ce13d87a

image

> Is it possible to add a "Manage" link that opens the sidebar panel?

Unfortunately, the notice only supports a string. We also have to give up on the icon for the linked post notice...

naxoc (approved):
Tested again and it works great. I'm also much happier with wording and the before-published behaviour 👍

includes/class-content-distribution.php
miguelpeixe: Whoops! Removed in 9a1516fbc11bc3064a66a0ae9c167b45660c57ab

miguelpeixe:
Thank you @naxoc, @thomasguillot, and @leogermani for your reviews here! 🙇

PR #138, opened 2024-09-09
Epic/network memberships handling

leogermani:
This epic branch adds handling of network subscriptions and memberships based on the data made available by https://github.com/Automattic/newspack-network/pull/122

It should be merged only once https://github.com/Automattic/newspack-network/pull/122 has been deployed and we have executed the backfills on every site.

The first PR in this epic branch reverts https://github.com/Automattic/newspack-network/pull/114. (merged)

###  1: https://github.com/Automattic/newspack-network/pull/135

This is a re-implementation of https://github.com/Automattic/newspack-network/pull/114

Testing:

* Create a granted-on-subscription memberships plan on two network sites, and make them synchronised (set same Network ID in the plan settings)
* On the Hub, buy a subscription tied to that membership plan as a reader
* On the Node, Pull all events
* Visit the Node as an anonymous reader and buy a subscription that gives access to the same membership plan using the same email
* Run  `wp newspack-network process-webhooks` in the node to sync events to the Hub
* On the Hub admin, Cancel the subscription
* Confirm the associated Membership is kept active, and that a note is added to the user membership saying it was kept active because there is an active subscription in the Node

###  4: https://github.com/Automattic/newspack-network/pull/137

This is a re-implementation of https://github.com/Automattic/newspack-network/pull/112

Testing: 

* Create a granted-on-subscription memberships plan on two network sites, and make them synchronised (set same Network ID in the plan settings)
* Make sure to sync the latest events by running  `wp newspack-network process-webhooks` and `wp newspack-network sync-all` in the Node 
* you should see the `newspack_network_membership_plan_updated` events in the Event log for both the plan in the hub and the plan in the node
* you also should see that the `newspack_network_membership_plans` option populated in both sites (this is what we tested in #124 
* One one site, buy the subscription, which should grant you the plan
* On the second site, log in/register using the same email address
* On the second site, try to buy the membership-granting subscription – observe it's not allowed

Captura de tela de 2024-08-07 12-42-47

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1207352955560208

PR #169, opened 2024-12-13
feat: limit purchase of a network membership

leogermani:
Limit subscription purchase if it grants access to a membership the user already has access to in another site of the Network.

This is the same thing as https://github.com/Automattic/newspack-network/pull/137 (instructions number 4 on https://github.com/Automattic/newspack-network/pull/138) but for logged out readers.

Here's how to test it:

* Having two sites in a Newspack Network installation
* On site A and B, create a membership plan with the same "Network ID"
* In each site, create a subscription product that grants access to these membership plans
* Make sure the network is in sync ( `wp newspack-network process-webhooks` in all Nodes and then `wp newspack-network sync-all` in all nodes)
* On site A, as a new reader, purchase the subscription
* Make sure the network is in sync again
* Visit site B, not logged in
* Attempt to purchase the subscription, using the same email used in site A
* You should not be allowed to make the purchase and should see a message saying that you already have that membership in another site:

image

---
  - 0-as-1207352955560208

### DIFF
--- a/includes/woocommerce-memberships/class-limit-purchase.php
+++ b/includes/woocommerce-memberships/class-limit-purchase.php
@@ -20,6 +20,10 @@ class Limit_Purchase {
 	public static function init() {
 		add_filter( 'woocommerce_subscription_is_purchasable', [ __CLASS__, 'restrict_network_subscriptions' ], 10, 2 );
 		add_filter( 'woocommerce_cart_product_cannot_be_purchased_message', [ __CLASS__, 'woocommerce_cart_product_cannot_be_purchased_message' ], 10, 2 );
+
+		// Also limit purchase for logged out users, inferring their IDs from the email.
+		// add_filter( 'woocommerce_subscriptions_product_limited_for_user', [ __CLASS__, 'subscriptions_product_limited_for_user' ], 10, 3 );
+		add_filter( 'woocommerce_cart_item_is_purchasable', [ __CLASS__, 'subscriptions_product_limited_for_user' ], 10, 4 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/woocommerce-memberships/class-limit-purchase.php
```diff
@@ -105,4 +115,48 @@ public static function woocommerce_cart_product_cannot_be_purchased_message( $me
...
+		$id_from_email = self::get_user_id_from_email();
+		if ( $id_from_email ) {
+			$cart_items = WC()->cart->get_cart();
+			foreach ( $cart_items as $cart_item ) {
+				$product = $cart_item['data'];
+				$network_active_subscription = self::get_network_equivalent_subscription_for_current_user( $product, $id_from_email );
+				if ( $network_active_subscription ) {
+					$error_message = sprintf(
+						/* translators: %s: Site URL */
+						__( "You can't buy this subscription because you already have it active on %s", 'newspack-network' ),
```
chickenn00dle: I'm fine with this updated error message, but just wanted to share why I removed the site name initially. Since this is for logged out readers, its possible for someone to input another persons email address here, in which case they'd learn about where another person's subscription exists.

This is probably not a big deal, but wanted to share just in case.

miguelpeixe (approved):
It works!

image

I agree with @chickenn00dle on the messaging here. I don't think it's great that we're disclosing that information but also think it's a very edge case.

leogermani:
@chickenn00dle @miguelpeixe , thanks for the reviews! I've updated the copy with help from @thomasguillot and @kmwilkerson to 

> Oops! You already have a subscription on another site in this network that grants you access to this site as well. Please log in using the same email address.

PR #170, opened 2024-12-13
feat(content-distribution): control distribution meta and prevent multiple dispatches

miguelpeixe:
Implements support for short-circuiting the distribution post meta update if it's not valid. Also implements support for distributing a post if a post meta changes, given we're also syncing them.

Because of the data event dispatch on post meta updates, there can be multiple dispatches in a single post save, if it includes multiple post meta updates. To prevent that, the distribution is now queued to be executed on `shutdown`. With this strategy, only the most recent payload will be distributed in the execution.

This PR also introduces a change to how the distribution meta for the outgoing post works. There's no need for the outgoing post configuration meta to include the network post ID, as it's deterministic. Without it, it's left only with the site URLs. Because of that, it's now only an array of site URLs, and the methods have been renamed from `set_config()` to `set_distribution()`.

### Testing

There should be no functional change, except `update_post_meta()` should now also trigger an update. Let's test it with the CLI:

1. Make sure you have a network setup with a hub and 2 nodes (node1 and node2)
2. In the hub, create a post and distribute to node1:

```
wp newspack network distribute post <post-id> --sites=https://node1.test
```

3. Confirm the post is distributed to node1
4. Distribute to node2 as well:

```
wp newspack network distribute post <post-id> --sites=https://node2.test
```

5. Confirm the post is distributed to node2
6. Now update a post meta:

```
wp post meta update <post-id> custom-meta test
```

7. Confirm the post meta gets distributed to both node1 and node2

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -161,6 +161,7 @@ class Content_Distribution {
 	 * @return void
 	 */
 	public static function handle_post_updated( $post ) {
+		$post = get_post( $post );
 		if ( ! $post ) {
 			return;
 		}
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -20,7 +20,7 @@ class Content_Distribution {
 	/**
 	 * Queued network post updates.
 	 *
-	 * @var array Map of post IDs to update.
+	 * @var array Post IDs to update.
 	 */
 	private static $queued_post_updates = [];
 
@@ -65,7 +65,7 @@ class Content_Distribution {
 	 * Distribute queued posts.
 	 */
 	public static function distribute_queued_posts() {
-		$post_ids = array_keys( self::$queued_post_updates );
+		$post_ids = array_unique( self::$queued_post_updates );
 		foreach ( $post_ids as $post_id ) {
 			$post = get_post( $post_id );
 			if ( ! $post ) {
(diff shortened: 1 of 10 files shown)


includes/class-content-distribution.php
```diff
@@ -17,31 +17,62 @@
...
-		Data_Events::register_listener( 'newspack_network_incoming_post_inserted', 'network_incoming_post_inserted', [ __CLASS__, 'handle_incoming_post_inserted' ] );
+		Data_Events::register_action( 'network_post_updated' );
+		Data_Events::register_action( 'network_incoming_post_inserted' );
+	}
+
+	/**
+	 * Distribute queued posts.
+	 */
+	public static function distribute_queued_posts() {
+		$post_ids = array_keys( self::$queued_post_updates );
```
naxoc: Since we don't check if the values are true or false in the `$queued_post_updates` array – what do you think about just holding the post ids instead of post_id -> bool in that array? Are there cases we'd be interested in false?

naxoc (commented):
Tests well! I added a small question/request

includes/class-content-distribution.php
miguelpeixe: I went for a map for efficiency as it's a simple way to ensure uniqueness.

miguelpeixe: Updated in f9eada237968c29c23b035e033fd05646daa7fc2

naxoc (approved):
Looks good! 
ok

PR #173, opened 2024-12-17
fix(content-distribution): post insertion hook and additional meta for incoming post event

miguelpeixe:
When rewriting the hooks for #170, I forgot to hook `handle_post_updated` to `wp_after_insert_post` to distribute on post content updates.

This PR adds the missing hook and also implements additional metadata for the post inserted event. These additional meta will be helpful for UI elements later on.

### Testing

Confirm a distributed post update reaches its connected nodes.

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -184,13 +184,16 @@ class Content_Distribution {
 			return;
 		}
 		$data = [
-			'origin'      => [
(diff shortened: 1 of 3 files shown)


naxoc (approved):
Tested and this looks good!

miguelpeixe:
Thank you for the review!

PR #175, opened 2024-12-18
fix: Force Distributor to not override bylines

leogermani:
Forces the Distributor option to override Bylines with the site URL to false and remove it from the UI.

image

### Testing

1. Visit Distributor > Settings
2. Make sure you don't see the "Override Author Byline" option
3. in a shell run `Distributor\Utils\get_settings();`
4. Confirm that `override_author_byline` is always set to `false`


chickenn00dle (approved):
Looks good!

PR #176, opened 2024-12-19
feat(content-distribution): capability and admin page

miguelpeixe:
Implements the capability to be managed in user roles for content distribution and the "Distribute" admin page.

The page should be rendered for both hub and node. In the hub, the additional "Node the Canonical URLs should point to" option was ported from the `Distributor_Settings` class and should behave the same.

The new page should take over `Distributor_Settings` if `NEWPACK_NETWORK_CONTENT_DISTRIBUTION` is on and the canonical URL option value should be applied to both implementations (Distributor customization and Newspack's Content Distribution).

This PR does not cover the canonical URL feature for content distribution but shouldn't break the Distributor customization.

| Hub | Node |
| --- | --- |
| image | image |

## Testing

1. In both hub and node, visit the new page at Newspack Network -> Distribute and confirm it renders as presented above
2. Confirm the canonical URL option still behaves as intended by following instructions from #1
3. Confirm the page renders with the default roles selected (admin, editor, and author)
4. Create a contributor user, draft a new post, and confirm you don't see the 🌐 icon in the editor
5. Back to the "Newspack Network -> Distribute" page, check "contributor", save and confirm the selection persisted
6. Draft a new post and confirm you see the 🌐 icon in the editor

### DIFF
--- a/includes/content-distribution/class-admin.php
+++ b/includes/content-distribution/class-admin.php
@@ -1,0 +1,320 @@
+<?php
+/**
+ * Newspack Network Content Distribution Admin Page.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network\Content_Distribution;
+
+use Newspack_Network\Content_Distribution;
+use Newspack_Network\Site_Role;
+use Newspack_Network\Hub\Nodes;
+use Newspack\Data_Events;
+use Newspack_Network\Admin as Network_Admin;
+
+/**
+ * Content Distribution Admin Page Class.
+ */
+class Admin {
+	/**
+	 * Capability to determine whether the user can distribute content.
+	 */
+	const CAPABILITY = 'newspack_network_distribute';
+
+	/**
+	 * The setting section constant
+	 */
+	const SETTINGS_SECTION = 'newspack_content_distribution_settings';
+
+	/**
+	 * The admin page slug
+	 */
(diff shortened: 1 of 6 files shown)


includes/content-distribution/class-admin.php
```diff
@@ -0,0 +1,321 @@
+<?php
+/**
+ * Newspack Network Content Distribution Admin Page.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network\Content_Distribution;
+
+use Newspack_Network\Content_Distribution;
```
naxoc: This one is unused

includes/content-distribution/class-admin.php
```diff
@@ -0,0 +1,321 @@
...
+	 * @param string $option The option name.
+	 * @return array
+	 */
+	public static function dispatch_canonical_url_updated_event( $old_value, $value, $option ) {
+		if ( '0' === (string) $value ) {
+			return [
+				'url' => get_bloginfo( 'url' ),
+			];
+		}
+		$node     = new Node( $value );
```
naxoc, line 311: We need a use statement for `Newspack_Network\Hub\Node`

naxoc (approved):
This tests well and works!

I added some very small non-blocking nits :)

miguelpeixe:
Thanks @naxoc! Fixed the imports in 68fd0530da9b6703ce01280f466461fca01f3d15

PR #177, opened 2024-12-20
feat(content-distribution): canonical url

miguelpeixe:
Implements canonical URL to content distribution. The default canonical URL should be the original poster of the content, but customizable via the "Node the Canonical URLs should point to" option in the hub's "Distribute" settings. It should behave exactly like #1.

Because the canonical URL filter will be called on every post page load, it warrants a performance tweak to the Incoming_Post class. Its constructor has been modified to not run a post meta query if the received parameter is a post ID.

This PR also reorganizes the tests a bit, introducing an `utils.php` file and placing them in their own directory and namespace.

### Testing

1. Confirm unit tests are accurate
2. Distribute a post and confirm in the node that the canonical URL metatag (search for "canonical" in the post page DOM) matches the original URL
3. In the hub, navigate to Newspack Network -> Distribute and change the "Node the Canonical URLs should point to" to a different site
4. Use CLI to propagate the option change in the node that's receiving the post: `wp newspack-network sync-all`
5. Visit the post page and confirm the canonical URL now matches the selected option

### DIFF
--- a/includes/class-accepted-actions.php
+++ b/includes/class-accepted-actions.php
@@ -41,6 +41,7 @@ class Accepted_Actions {
 		'network_nodes_synced'                     => 'Nodes_Synced',
 		'newspack_network_membership_plan_updated' => 'Membership_Plan_Updated',
 		'network_post_updated'                     => 'Network_Post_Updated',
+		'network_post_deleted'                     => 'Network_Post_Deleted',
 	];
 
 	/**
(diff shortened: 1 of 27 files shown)


naxoc (approved):
Tested and eyeballed. Looks good!

miguelpeixe:
Thanks, @naxoc!

PR #179, opened 2024-12-23
feat(content-distribution): sync comment and ping statuses

miguelpeixe:
Implements synchronization of comment and ping statuses on distribution.

### Testing

1. Double-check the unit test is accurate
2. Distribute a post and, after distribution, change the comment and ping statuses to `closed` and save
3. Confirm in the destination post that both statuses have been updated to `closed`


naxoc (approved):
Looks good! 👍

PR #180, opened 2024-12-23
feat(event-log): collapse data

miguelpeixe:
The content distribution project introduces events with significantly larger payloads shared between network sites. Because of that, we need better handling of our event log table.

This PR proposes collapsing the data contents inside a disabled textarea with a "Copy to clipboard" button. The idea is to suggest parsing that information somewhere else, which will always be more suitable than any UI we could produce here.

image

### Testing

1. Visit the Event Log page in the hub
2. Clip "Copy to clipboard" on any item
3. Confirm it pastes the contents


naxoc (approved):
Works great! Nice and elegant solution to not enough space 👍

miguelpeixe:
Thanks, @naxoc! Since this changes a broader plugin feature, I'll hold the merge until @leogermani's feedback.

leogermani:
What do you think of doing this only if the data length is larger than XX chars? It is handy to be able to read the small events directly on the table

miguelpeixe:
Good call. Changed in 1ce9d729411f6973055ba9703dbea2c49e8c063d with a 300 characters threshold.

image

leogermani (approved):
LGTM

PR #181, opened 2025-01-03
feat(content-distribution): Block the editor for incoming posts

naxoc:
This adds the UI editor "blocking" for incoming posts. 

The editor is disabled and save is not possible (from the UI). I haven't put any protections in place in PHP on save by API calls or similar. I can add that if we think it's necessary.

Note that I changed the text from "This post is linked to the origin post. Edits to the origin post will update this **remote** version" to "This post is linked to the origin post. Edits to the origin post will update this version" in the sidebar for linked posts.

I added the function `get_original_site_url()` – I stole it from #177 :)


src/content-distribution/incoming-post/index.js
```diff
@@ -0,0 +1,87 @@
...
+		>
+			<Panel>
+				<PanelBody className="distribute-header">
+					{isLinked
+						? __(
+								'This post is linked to the origin post. Edits to the origin post will update this version',
+								'newspack-network'
+							)
+						: __(
+								'This post has been unlinked from the origin post. Edits to the origin post will not update this version',
```
miguelpeixe: NIT:

```suggestion
								'This post is linked to the origin post. Edits to the origin post will update this version.',
								'newspack-network'
							)
						: __(
								'This post has been unlinked from the origin post. Edits to the origin post will not update this version.',
```

src/content-distribution/incoming-post/index.js
```diff
@@ -0,0 +1,87 @@
...
+
+function IncomingPost() {
+	const { createNotice } = useDispatch('core/notices');
+	const { lockPostSaving, lockPostAutosaving } = useDispatch('core/editor');
+	const { openGeneralSidebar } = useDispatch('core/edit-post');
+
+	useEffect(() => {
+		lockPostSaving('distributed-incoming-post-lock');
+		lockPostAutosaving('distributed-incoming-post-lock');
+	});
```
miguelpeixe: This should probably be conditioned to `isLinked`

miguelpeixe (changes requested):
Good solution blocking the block editor!

When editing a linked post, I'm still able to use the side panel forms:

image

These should either render disabled inputs or nothing at all. Not sure which route, WDYT?

When a distributed post comes in, it's expectedly a draft. When trying to publish, the button is disabled, possibly because of the `lockPostSaving()`. We still need to allow a post to be published (or any status change), can we find a way around that?

image

When a post gets unlinked, I'm still getting the editor blocked. I should be allowed to make changes when unlinking a post.

I'm also missing the ability to unlink/link. Will it be tackled separately? It might be convenient to handle in the same PR, since that change has a direct impact in how this UI should behave.

naxoc:
Thanks for the review!

I realized I am missing quite a few things here, so I'm back at the drawing board :) 

I'll add/fix the things you are pointing out

naxoc:
OK, here is take two.

The CSS is still all that enforces the "blocking". Parts of it became a bit more fragile than I like, so I'm open to suggestions.

I added a component `distribute-panel` that we could re-use for the distribute panel for what is in the JS `distribute` component now (and maybe rename that to outgoing-post), but we can do that in a separate PR also.

How to test:
* Distribute a post and play around with the UI.
* Toggle the post linked and unlinked. A linked post should lock down:
  * The editor
  * The sidebar (execpt for the one that has the status, trash button, etc).
  * The pre-publish sidebar – except the bits we allow (please review that CSS and see if that is too fragile)
  * The bottom part of the editor where Yoast lives for instance.

src/content-distribution/incoming-post/index.js
```diff
@@ -0,0 +1,135 @@
...
+					<Button
+						variant={ isUnLinked ? 'primary' : 'secondary' }
+						isDestructive={ !isUnLinked }
+						disabled={ isLinkedToggling }
+						onClick={ () => {
+							toggleLinked();
+						} }
+					>
+						{ !isUnLinked ? __( 'Unlink from origin post', 'newspack-network' ) : __( 'Relink to origin post', 'newspack-network' ) }
+					</Button>
```
miguelpeixe: It'd be nice to have the `isBusy` state here when it's updating, with the label updated to "Linking..." or "Unlinking...", [like we have for distributing](https://github.com/Automattic/newspack-network/blob/2d676c2c0da66bdac0d6c9b1f342d5c51d3dc657/src/content-distribution/distribute/index.js#L217-L234).

src/content-distribution/incoming-post/index.js
```diff
@@ -0,0 +1,135 @@
...
+	useEffect( () => {
+		setIsUnLinked( unlinked );
+	}, [ unlinked ] );
+
+	useEffect( () => {
+		setTimeout( () => {
+			openGeneralSidebar(
+				'newspack-network-incoming-post/newspack-network-distribute-panel'
+			);
+		}, 10 ); // TODO. There must be a better way
```
miguelpeixe: I found something slightly better. The hook dependency can be [`areMetaBoxesInitialized`](https://developer.wordpress.org/block-editor/reference-guides/data/data-core-edit-post/#aremetaboxesinitialized). Even though I'm not sure this initialization is directly tied to the sidebar, it consistently worked for me and it feels less arbitrary than a timeout.

It can be selected from the `core/edit-post` store:

```js
	const { postId, areMetaBoxesInitialized } = useSelect( select => {
		const {
			getCurrentPostId,
		} = select( 'core/editor' );
		const {
			areMetaBoxesInitialized,
		} = select( 'core/edit-post' );
		return {
			postId: getCurrentPostId(),
			areMetaBoxesInitialized: areMetaBoxesInitialized(),
		};
	} );

	useEffect( () => {
		if ( areMetaBoxesInitialized ) {
			openGeneralSidebar(
				'newspack-network-incoming-post/newspack-network-distribute-panel'
			);
		}
	}, [ areMetaBoxesInitialized ] );
```

miguelpeixe (changes requested):
Thank you for the revisions! It's looking great!

> I added a component distribute-panel that we could re-use for the distribute panel for what is in the JS distribute component now (and maybe rename that to outgoing-post), but we can do that in a separate PR also.

I loved that idea! Let's do this after this gets merged.

I left a couple of comments below, but also wanted your thoughts on the relinking action. When a post is unlinked, you're able to make changes to it and relinking will overwrite those changes. On the backend this is working well, but on the editor you'll only see the resynced content after a page refresh, not right after relinking is complete.

Ideally, we'd have that content resync instantly, but I'm not familiar with a way to make Gutenberg pull content and refresh its store from the backend. Do you have ideas?

src/content-distribution/distribute-panel/index.js
```diff
@@ -0,0 +1,40 @@
...
+import { PluginSidebar } from '@wordpress/editor';
+import { Flex, Panel, PanelBody } from '@wordpress/components';
+import { globe } from '@wordpress/icons';
+
+/**
+ * Internal dependencies.
+ */
+import './style.scss';
+
+const DistributePanel = ({ header, body, footer, buttons }) => {
```
miguelpeixe: Since we have the `Distribute()` component/plugin, maybe we can call this `ContentDistributionPanel`?

naxoc: Yeah, that makes sense. Unless you think renaming the `Distribute()` component to `OutgoingPost` to match the backend code and this one?

miguelpeixe: > Unless you think renaming the Distribute() component to OutgoingPost to match the backend code and this one?

Hmm, I might. Either way, I think `ContentDistributionPanel` is more suitable.

src/content-distribution/incoming-post/index.js
naxoc: A hundred times better. Thanks!

naxoc:
I re-did things a bit and spent a bunch of time researching whether it's possible to replace all content in the editor on relinking. The post content and title is easy enough, but I could find no way to guarantee that I would not overwrite the tags, categories, or whatever else was edited, so I went with just reloading the page on relinking. That created its own problems to avoid the browser dialogue with "You have unsaved changes" so I have quite a lot of async chaining in the code to work around that. Suggestions welcome to make that more sleek!

I also renamed the panel component and fixed up the CSS a bit more.

miguelpeixe (approved):
Thank you for the revisions, @naxoc! It's working well and the UI is solid.

I found 2 things that we can tackle in another PR:

- When the post is linked (UI blocked), if I navigate to the post settings panel and click on the "Block" tab, I can't click back to the "Post" tab. Something about the UI blocking strategy takes over that area as well.
- An oversight on my part, we should also block the ability to change the featured image:

Captura de Tela 2025-01-16 às 09 38 38

PR #185, opened 2025-01-14
feat(content-distribution): migrator

miguelpeixe:
Implements a migration tool so sites using Distributor can move to our content distribution functionality for posts being distributed within the network.

The migration is done via CLI:

```
NAME

  wp newspack network distributor migrate

DESCRIPTION

  Migrate posts from Distributor to Newspack Network's content distribution

SYNOPSIS

  wp newspack network distributor migrate [<post-id>] [--batch-size=<batch-size>]
  [--all] [--strict] [--delete]

OPTIONS

  [<post-id>]
    The ID of the post to migrate.

  [--batch-size=<batch-size>]
    Number of posts to migrate in each batch.
    ---
    default: 50
    ---

  [--all]
    Migrate all posts.

  [--strict]
    Whether to only migrate if all distributed posts can be migrated.

  [--delete]
    Whether to deactivate and delete the Distributor plugin after migrating all
    posts. This will only take effect if all posts were able to migrate.

```

The tool migrates Distributor's subscriptions, using its metadata to create an outgoing post. A data event is dispatched to migrate the incoming posts so this is handled asynchronously. Because of that, it runs in batches of 50.

### Testing

1. In your network setup, make sure you have Distributor configured to all sites in the network
2. Create a new site, outside the network, and add it to your Distributor connections in order to test an attempt to migrate content distributed to a site outside the network
3. Publish posts with the following Distributor distribution:
    1. to 1 site within the network
    2. to 1 site outside the network
    3. to 2 sites within the network
    4. to all sites (within and outside the network)
4. Run the CLI to migrate the first post:

```
wp newspack network distributor migrate {post_id}
```

5. Visit the post origin and confirm the 🌐 panel reflects the correct setup
6. Confirm the post no longer has "Distributor" settings
7. Make a change and save
8. Make sure the network events are synced in all nodes: `wp newspack-network sync-all`
9. Confirm the change is distributed via the network's content distribution
10. Run the CLI to migrate all posts in strict mode:

```
wp newspack network distributor migrate --all --strict
```

11. Confirm no posts were migrated
12. Run the CLI to migrate all posts:

```
wp newspack network distributor migrate --all
```

13. Confirm that the posts with distribution outside the network were not migrated
14. For the posts that were able to migrate, repeat steps 5-8 for each post
15. Delete the posts that are being distributed outside the network and publish a new post distributed, via Distributor, within the network
16. Run the CLI again, but now flagging the plugin deletion:

```
wp newspack network distributor migrate --all --delete
```

17. Confirm the plugin has been deleted after the migration


includes/content-distribution/class-distributor-migrator.php
```diff
@@ -0,0 +1,234 @@
...
+					'subscription_signature' => [
+						'type'     => 'string',
+						'required' => true,
+					],
+					'network_post_id'        => [
+						'type'     => 'string',
+						'required' => true,
+					],
+				],
+				'permission_callback' => '__return_true', // TODO: Check network signature.
```
miguelpeixe: To complete the migration, there has to be a request from the distribution origin to its destination in order to set the network post ID, otherwise the migration will result in a duplicated post.

To secure that request, do we already have a single way that supports node->hub, hub->node, and node->node requests? cc @leogermani

miguelpeixe: FTR, this endpoint is currently secured by Distributor's subscription signature, which is how Distributor secures it. We can also decide that this is enough security for this endpoint.

includes/content-distribution/class-distributor-migrator.php
```diff
@@ -0,0 +1,392 @@
...
+class Distributor_Migrator {
+	/**
+	 * Initialize hooks.
+	 */
+	public static function init() {
+		add_action( 'init', [ __CLASS__, 'register_data_event_actions' ] );
+	}
+
+	/**
+	 * Register data event ations.
```
naxoc: Tiny typo in actions

includes/content-distribution/class-distributor-migrator.php
```diff
@@ -0,0 +1,392 @@
...
+/**
+ * Newspack Network Content Distribution Distributor Migrator.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network\Content_Distribution;
+
+use Newspack\Data_Events;
+use Newspack_Network\Content_Distribution;
```
naxoc: These use statements are not used:
```
use Newspack_Network\Content_Distribution;
use WP_Query;
use WP_Post;use InvalidArgumentException;
use WP_REST_Request;
use WP_REST_Response;
```

includes/content-distribution/class-distributor-migrator.php
```diff
@@ -0,0 +1,392 @@
...
+
+	/**
+	 * Migrate a post from Distributor to Newspack Network Content Distribution.
+	 *
+	 * @param int  $post_id                The ID of the post to migrate.
+	 * @param bool $migrate_incoming_posts Whether to migrate incoming posts.
+	 *
+	 * @return Outgoing_Post|WP_Error Outgoing_Post on success, WP_Error on failure.
+	 */
+	public static function migrate_outgoing_post( $post_id, $migrate_incoming_posts = true ) {
```
naxoc, line 267: It looks like this function is unused?

includes/content-distribution/class-distributor-migrator.php
```diff
@@ -0,0 +1,392 @@
...
+		$can_migrate = self::can_migrate_outgoing_post( $post_id );
+		if ( is_wp_error( $can_migrate ) ) {
+			return $can_migrate;
+		}
+
+		$subscriptions = get_post_meta( $post_id, 'dt_subscriptions', true );
+
+		$outgoing_post = null;
+		foreach ( $subscriptions as $subscription_id ) { // phpcs:ignore WordPressVIPMinimum.Functions.CheckReturnValue.NonCheckedVariable
+			$migration_result = self::migrate_subscription( $subscription_id, $migrate_incoming_post );
```
naxoc: The `$migrate_incoming_post` var is uninitialised

includes/content-distribution/class-cli.php
```diff
@@ -107,4 +150,91 @@ public function cmd_distribute_post( array $pos_args, array $assoc_args ): void
...
+			$batches    = array_chunk( $post_ids, $batch_size );
+
+			foreach ( $batches as $i => $batch ) {
+				$result = Distributor_Migrator::migrate_outgoing_posts( $batch );
+				if ( is_wp_error( $result ) ) {
+					$message = sprintf( '(%d/%d) Error migrating batch: %s', $i + 1, count( $batches ), $result->get_error_message() );
+					if ( $strict ) {
+						WP_CLI::error( $message );
+					} else {
+						$errors->add( $result );
```
naxoc: Required param is missing here - probably the error code?

naxoc (changes requested):

naxoc (changes requested):
Found a couple of smaller things in code – will do a test next, but m not sure I entirely get step 3. 

Are these assumptions correct?

Let's say I have hub.test, node1.test, node2.test, node3.test.
hub, node1, and node2 are networked (content distribution enabled). node3 is not. All sites have distributor enabled and are networked.

* to 1 site within the network - so post on hub and distribute to node1?
* to 1 site outside the network - so post on hub and distribute to node3?
* to 2 sites within the network - so post on hub and distribute to node1 and node2?
* to all sites (within and outside the network) - so post on hub and distribute to node1, node2, and node3? (I guess this one is not so hard to understand).

Am I right about the above?

includes/content-distribution/class-distributor-migrator.php
miguelpeixe: It's meant to be used with the CLI when migrating a single post, but the name of the method was wrong. Fixed in 893ed4e898869b8acb5199531315639382ed96e6

includes/content-distribution/class-distributor-migrator.php
miguelpeixe: Fixed in ecd3d704062842eb5d6448625f97fd15ce3a3a76

includes/content-distribution/class-distributor-migrator.php
miguelpeixe: Fixed in 0bfa9a733e4d1d404c8fda0f59885980e08fb56c

includes/content-distribution/class-distributor-migrator.php
miguelpeixe: Fixed in ad8fdd53749132553019f000a13ab80168bb9c1b

includes/content-distribution/class-cli.php
miguelpeixe: Fixed in 1237389b119cded6183b43dec87706cceb7b2ac8

miguelpeixe:
Thanks for the review, @naxoc!

> Am I right about the above?

Yes, that's the setup for testing.

A few things were missing to handle the incoming posts migration, which have been implemented in 1d5f4d17e9be818803b6bc783cd7df371c93994d and improved logging was also added in 6f7af2c823a8762322f18f506700d9cfeb373cf0

naxoc (approved):
👍 🎉 

I ran through the testing scenario and it works like a charm now.

PR #186, opened 2025-01-15
fix(content-distribution): persist site hash

miguelpeixe:
We're not expecting a site URL to change, which is expected to break their network communications (not just content distribution). Nonetheless, I've been having some issues with the network post ID because my local setup has a dynamic site URL, generating different IDs whether the code is running via shell, CLI, or the browser.

This change is to ensure that the site hash is consistent and the relationship between distributed posts never breaks.

## Testing

Distribution should work the same. Distribute a post to a node, save a few changes, and confirm the post (and its changes) are distributed.


naxoc (approved):
Good one! We would run into this when going from staging to live for instance, so nice catch.

PR #187, opened 2025-01-15
feat: sync multiple user roles

laurelfulford:
This PR updates the manual syncing to sync multiple user roles to support plugins like Members.

See: 1200550061930446-as-1209086108652380

### Steps to test:

1. Download and install Members (ping me if you need a copy!).
2. Set up some users with more than one role.
3. Try manually syncing those users to other sites using the "Sync user across network" button on each profile.
4. Note that only one role is sent, and applied to the reader when synced.
5. Apply this PR.
6. Repeat manually syncing users; confirm that multiple roles are sent in the logs and applied to the user.
7. Try a few variations of changes, including only adding or only removing roles, and completely switching the set of roles applied to the reader.
8. Disable Members and repeat testing syncing user roles of existing users. Note that if you sync a user at this point without resaving them, multiple roles will still be assigned to them. You'll need to resave the user's profile to get them back down to the default one role.
9. Smoke test creating new users, changing their roles, and syncing them across the network to make sure things still work as expected when Members is not active.


leogermani (approved):
Works as described!

laurelfulford:
Thanks @leogermani!

PR #189, opened 2025-01-17
feat(content-distribution): post status on create

miguelpeixe:
1208510338230630-as-1209168371395424
p1737121608239039-slack-C055SRT7WN4

Support configuring the post status on distribution. This should only take effect when creating the post.

image

image

image

### Testing

1. Distribute a post without changing the status and confirm it's distributed as a draft.
2. Distribute a post changing the status to "Published" and confirm it's distributed as published.
3. Do the same via CLI and confirm it also reflects the status:

```
newspack network distribute post {post_id} --sites=https://node1.local --status_on_create=publish
```


src/content-distribution/incoming-post/style.scss
```diff
@@ -28,19 +28,16 @@
 	}
 
 	.editor-sidebar:not(#newspack-network-incoming-post\:newspack-network-content-distribution-panel, #edit-post\:document) {
+		@include grey-out-outer;
```
naxoc: Looks like this one might also overwrite  #193

src/icons/drafts.js
```diff
@@ -0,0 +1,16 @@
+/**
```
naxoc, line 1: Very minor: but why is it drafts plural? Should it not be singular?

naxoc (approved):
Looks good! 👍 

Added a couple of small comments, but feel free to merge once you are done with those :)

src/icons/drafts.js
miguelpeixe: I didn't want to diverge from the source: https://github.com/WordPress/gutenberg/blob/2943dd3adfa4e1ad33b980760277377df9ac0f91/packages/editor/src/components/post-status/index.js

src/content-distribution/incoming-post/style.scss
miguelpeixe: 🤦  I did a very bad job fixing conflicts, sorry about that!

Fixed in 8ed2220456cc0e3c29bd5a60c68c7bb063684355

src/icons/drafts.js
naxoc: That makes sense. Thanks!

PR #192, opened 2025-01-17
fix(memberships): remove managed flag on cancel or expire

chickenn00dle:
Closes 1200550061930446-as-1209019125448157/f

This PR removes the network managed meta flag when a network membership is cancelled or expired, and resets network membership meta otherwise. This is so readers can resume a cancelled or expired subscription on another network site.

Right now, if a reader has an inactive membership on one site, then tries to rejoin the membership on another site, the managed meta flag and other meta data causes a plan mixup between nodes as in the asana task above.

**Steps to test**

1. As a reader fulfill the requirements to join a membership on a site on the network
2. Either wait for the membership to propogate, or force via the `wp newspack-network sync-all` tool.
3. Verify the membership on another site on the network
4. On the site where the membership is managed, cancel the membership fully, either as reader or admin and wait for event to propogate
5. On the second network site, verify the cancelled status of the membership
6. On the second site again, log in as the same reader, and fulfill the requirements to join the same membership on this site this time, then wait for the event to propogate
7. On the first site, verify the membership is now managed by the second site


leogermani (changes requested):
Interesting! So you're saying that if the membership is being permanently cancelled (not on hold) we should unlink them from the network, allowing it to be reactivated by a subscription purchase in another site. Makes sense!

Looking at the code, looks like the 3 `update_post_meta` starting on line 87 can be deleted after your changes (and it makes sense, because we don't check if `$user_membership` is a good object at that point)

I think it's worth adding a slightly bigger comment block explaining this reasoning there, so we can remember why we do this.

Not sure if it's also worth to add a note to the Membership saying it was unlinked from the other site?

chickenn00dle:
Thanks for the review @leogermani!

I've updated the PR with some more thorough comments, added a membership note, and am deleting all managed meta in https://github.com/Automattic/newspack-network/pull/192/commits/8e8a92ebafb3c9dc03757fc93708e22519f96c5e

Let me know if these comments need to be more clear!

includes/incoming-events/class-woocommerce-membership-updated.php
```diff
@@ -77,13 +77,13 @@ public function update_membership() {
...
 		if ( null === $user_membership ) {
+			// Create a new membership if it doesn't exist and link it to the remote membership.
 			$user_membership = wc_memberships_create_user_membership(
 				[
 					'plan_id' => $local_plan_id,
 					'user_id' => $user->ID,
 				]
 			);
-
 			update_post_meta( $user_membership->get_id(), Memberships_Admin::NETWORK_MANAGED_META_KEY, true );
```
leogermani: I mean you can delete these 3 lines

chickenn00dle: Ah whoops! Misread that. Updated in https://github.com/Automattic/newspack-network/pull/192/commits/b0678ebd08c4685b06e11417cd93838bd64a9f5f

chickenn00dle: Sorry. Also snuck in a change to one of the comments relevant here: https://github.com/Automattic/newspack-network/pull/192/commits/299a9ca153bfa95eedc376e71a0872bfa797fe01

leogermani (approved):
Tests well!

PR #193, opened 2025-01-20
fix(content-distribution): Improve CSS for blocked editor

naxoc:
This fixes 2 things that came out of #181

* Also blocks the "Featured image" button so that is not editable on linked posts
* Fixes a bug where it was not possible to click from the "Block" tab in the sidebar to the "Post" tab.

## How to test
On a linked post – ensure that the sidebar blocks what it should (and no more) and that you can click between the tabs in the sidebar.


miguelpeixe (approved):
Works well!

PR #194, opened 2025-01-20
feat(content-distribution): Sync authors

naxoc:
This copies most of the existing code for distributed post authors.

I tried to make the naming of classes make a bit more sense (using outgoing/incoming) naming-wise, but I think I might have lost my way somewhere in the process. It's still pretty unwieldy. Suggestions very welcome!

One thing I still lack: when a post is new and I click "publish" and then distribute it (without having touched authors), the cap authors are not added to the payload. I think the best way to fix that is to save the post from JS before distributing the first time. What do you think @miguelpeixe ?

## How to test
* Try distributing a post without having CAP enabled. Check that the author is transferred in a way that makes sense
* Try with an author that is a guest contributor. Also try mixing "normal" users and guest contributors.
* Try with Guest Authors enabled. Try mixing here too.

Still remains:
* Write more tests, but I want to use #203 to that when this is is merged if that is OK.


includes/content-distribution/class-cap-authors.php
```diff
@@ -0,0 +1,196 @@
...
+		}
+		// If the terms are the same, we don't need to do anything. Note that one array has string values and one has
+		// int values, so we use array_map with the intval for the comparison.
+		if ( array_map( 'intval', $old_tt_ids ) === array_map( 'intval', $tt_ids ) ) {
+			return;
+		}
+
+		try {
+			$outgoing_post = new Outgoing_Post( $object_id );
+			if ( ! $outgoing_post->is_distributed() ) {
```
naxoc: I don't really want to do this on all posts, but if I remove these lines the author change works a bit more consistently.

miguelpeixe: You can use this:

https://github.com/Automattic/newspack-network/blob/22240db138479453eadabd208a05756c0068466f/includes/class-content-distribution.php#L316-L325

includes/content-distribution/class-outgoing-post.php
```diff
@@ -211,6 +215,7 @@ public function get_payload( $status_on_create = 'draft' ) {
 			'status_on_create' => $status_on_create,
 			'post_data'        => [
 				'title'          => html_entity_decode( get_the_title( $this->post->ID ), ENT_QUOTES, get_bloginfo( 'charset' ) ),
+				'author'         => $post_author,
```
miguelpeixe: `Outgoing_Author::get_wp_user_for_distribution` can return `WP_Error`. We shouldn't send that to the payload.

includes/content-distribution/class-cap-authors.php
naxoc: That makes sense. I changed the code now to call distribute_partial so I need the outgoing post now.

includes/class-content-distribution.php
```diff
@@ -47,7 +49,7 @@ public static function init() {
 		}
 
 		add_action( 'init', [ __CLASS__, 'register_data_event_actions' ] );
-		add_action( 'shutdown', [ __CLASS__, 'distribute_queued_posts' ] );
+		add_action( 'shutdown', [ __CLASS__, 'distribute_queued_posts' ], 20 );
```
miguelpeixe: Any reason for the change in priority?

includes/content-distribution/class-cap-authors.php
```diff
@@ -0,0 +1,166 @@
...
+				return;
+			}
+
+			Content_Distribution_Class::distribute_post_partial( $outgoing_post->get_post(), [ 'multiple_authors' ] );
```
miguelpeixe:
```suggestion
			Content_Distribution_Class::queue_post_distribution( $object_id, 'multiple_authors' );
```

Although this should work well, we can use the `queue_post_distribution()` here. If the same request runs other distribution requests for the same post, it'll run only once on shutdown.

The try/catch here is also unnecessary. If we stick with the current approach, `distribute_post_partial()` accepts a post ID integer and the `is_post_distributed()` method seems more suitable here.

includes/content-distribution/class-incoming-post.php
```diff
@@ -584,9 +607,9 @@ public function insert( $payload = [] ) {
 		/**
 		 * Fires after an incoming post is inserted.
 		 *
-		 * @param int   $post_id   The post ID.
-		 * @param bool  $is_linked Whether the post is linked.
-		 * @param array $payload   The post payload.
+		 * @param int $post_id The post ID.
+		 * @param bool $is_linked Whether the post is linked.
+		 * @param array $payload The post payload.
```
miguelpeixe: Why this change in linting?

includes/content-distribution/class-outgoing-post.php
```diff
@@ -202,15 +202,20 @@ public function get_payload_hash( $payload = null ) {
...
+		$multiple_authors = apply_filters( 'newspack_network_multiple_authors_for_post', [], $this->post );
+
 		return [
 			'site_url'         => get_bloginfo( 'url' ),
 			'post_id'          => $this->post->ID,
 			'post_url'         => get_permalink( $this->post->ID ),
 			'network_post_id'  => $this->get_network_post_id(),
 			'sites'            => $this->get_distribution(),
 			'status_on_create' => $status_on_create,
+			'multiple_authors' => is_wp_error( $multiple_authors ) ? [] : $multiple_authors,
```
miguelpeixe: Partial payload filters through the contents of `post_data`. This should move there in order to work as designed.

As it is now, at the root-level, it'll be always part of the payload. It works, but not how we expect it to.

includes/content-distribution/class-outgoing-author.php
```diff
@@ -0,0 +1,63 @@
...
+class Outgoing_Author {
+
+	/**
+	 * Gets the user data of a WP user to be distributed along with the post.
+	 *
+	 * @param int|WP_Post $user The user ID or object.
+	 *
+	 * @return WP_Error|array
+	 */
+	public static function get_wp_user_for_distribution( $user ) {
```
miguelpeixe: I don't think we need a whole class for this. WDYT of moving this to the `Outgoing_Post` class?

It's similar to other methods we have there that are meant to build the payload, like `get_processed_post_content()`, `get_post_taxonomy_terms()`, and `get_post_meta()`.

includes/content-distribution/class-incoming-author.php
```diff
@@ -0,0 +1,86 @@
...
+namespace Newspack_Network\Content_Distribution;
+
+use Newspack_Network\Debugger;
+use Newspack_Network\User_Update_Watcher;
+use Newspack_Network\Utils\Users as User_Utils;
+
+/**
+ * Class to handle author ingestion for content distribution.
+ */
+class Incoming_Author {
```
miguelpeixe: I also think this class could become methods inside `Incoming_Post`.

includes/content-distribution/class-incoming-post.php
```diff
@@ -546,19 +560,28 @@ public function insert( $payload = [] ) {
...
 				self::log( 'Failed to insert post.' );
+
 				return new WP_Error( 'insert_error', __( 'Error inserting post.', 'newspack-network' ) );
 			}
 
 			// Update the object.
 			$this->ID   = $post_id;
 			$this->post = get_post( $this->ID );
 
+			Incoming_Author::ingest_author_for_post( $this->ID, $this->get_original_site_url(), $post_data['author'] );
```
miguelpeixe: This method makes a second post update to the db.  Can we dissolve it and move its logic upper in this method's `$postarr` arguments?

Regular WP author is a core aspect of the post and should fit well in the composition of our insert() logic.

includes/content-distribution/class-outgoing-post.php
```diff
@@ -202,15 +202,20 @@ public function get_payload_hash( $payload = null ) {
 	 * @return array|WP_Error The post payload or WP_Error if the post is invalid.
 	 */
 	public function get_payload( $status_on_create = 'draft' ) {
+		$post_author = $this->post->post_author ? Outgoing_Author::get_wp_user_for_distribution( $this->post->post_author ) : [];
+		$multiple_authors = apply_filters( 'newspack_network_multiple_authors_for_post', [], $this->post );
```
miguelpeixe: Can you share more about the use case for this filter? It's missing a docblock.

miguelpeixe (changes requested):
Thank you for refactoring the approach to fit the new partial payload implementation!

I'm able to get multiple authors synced, but with the standard guest author approach from CAP I'm unable to get the guest authors distributed.

I'm activating it via `define( 'NEWSPACK_ENABLE_CAP_GUEST_AUTHORS', true );` on all sites. Should I be doing something different?

includes/content-distribution/class-incoming-post.php
naxoc: It was probably just my editor when it autoformats. There are too many spaces I think it picked up?

includes/class-content-distribution.php
naxoc: No, I think that is a relic from when I was trying to game the timing of things. :) I removed it

includes/content-distribution/class-outgoing-post.php
naxoc: It was an attempt to make it so other implementations of multiple authors (like other than the CAP plugin) could chime in. I'm missing an equivalent on incoming – I'll add that and add those in-function docblocks

includes/content-distribution/class-incoming-post.php
miguelpeixe: It follows the standard we use for functions and methods so the parameters info is aligned. It could be that your IDE doesn't interpret the `do_action()` as a thing that can precede that type of docblock.

naxoc: You are right! I had the editor set to not WP for some reason and it felt like I had to keep tweaking it :)

includes/content-distribution/class-outgoing-post.php
naxoc: 👍 Moved it to post_data

includes/content-distribution/class-outgoing-post.php
naxoc: I also added a filter for incoming "mulitple_authors", so any implementation could hook in in both directions.

includes/content-distribution/class-incoming-post.php
naxoc: Makes sense! thanks :)

naxoc:
Thanks for the review @miguelpeixe

I implemented your suggestions. If you think the whole hooks thing for multiple authors is over engineering things, let me know and I can change them to function calls.

includes/content-distribution/class-cap-authors.php
```diff
@@ -0,0 +1,160 @@
...
+	public static function is_co_authors_plus_active(): bool {
+		global $coauthors_plus;
+
+		return $coauthors_plus instanceof \CoAuthors_Plus && function_exists( 'get_coauthors' );
+	}
+
+	/**
+	 * Action callback.
+	 *
+	 * Add a postmeta entry with the Co-Authors Plus authors for outgoing posts.
```
miguelpeixe: This docblock needs an update.

includes/content-distribution/class-outgoing-post.php
miguelpeixe: Ah, I see. I didn't realize that's how the CAP data is being added to the payload.

Non-blocking suggestion: maybe instead of filtering this specific property we can go with a more generic approach and filter the payload `post_data` to let additional features determine their properties. That'll also generate a cleaner payload for sites that don't have CAP at all - no permanent empty array.

miguelpeixe (changes requested):
Thank you for the revisions, @naxoc! It's looking great.

Maybe you missed this bit from my previous review:

> I'm able to get multiple authors synced, but with the standard guest author approach from CAP I'm unable to get the guest authors distributed.
> 
> I'm activating it via define( 'NEWSPACK_ENABLE_CAP_GUEST_AUTHORS', true ); on all sites. Should I be doing something different?

I'm still unable to get it to work. The payload does include the data, though:

```
{
    "site_url": "http:\/\/localhost",
    "post_id": 827,
    "post_url": "http:\/\/localhost\/2025\/01\/my-new-cap-post\/",
    "network_post_id": "6dc7943077f7903f656015af0ebb029e",
    "sites": [
        "http:\/\/node1.test"
    ],
    "status_on_create": "draft",
    "post_data": {
        "multiple_authors": [
            {
                "type": "wp_user",
                "ID": 1,
                "display_name": "wordpress",
                "user_email": "wordpress@example.com",
                "user_url": "http:\/\/localhost"
            },
            {
                "ID": "825",
                "display_name": "My CAP Guest Author",
                "first_name": "",
                "last_name": "",
                "user_login": "my-cap-guest-author",
                "user_email": "my-cap-guest-author@example.com",
                "linked_account": "",
                "website": "",
                "description": "",
                "newspack_job_title": "",
                "newspack_role": "",
                "newspack_employer": "",
                "newspack_phone_number": "",
                "user_nicename": "my-cap-guest-author",
                "type": "guest_author",
                "nickname": ""
            },
            {
                "type": "wp_user",
                "ID": 2,
                "display_name": "Contributor",
                "user_email": "contributor@example.com",
                "first_name": "Contributor",
                "last_name": "Contributor"
            }
        ],
        "date_gmt": "2025-01-30 17:37:01",
        "modified_gmt": "2025-01-30 17:39:22"
    },
    "partial": true
}
```

naxoc:
Woops. I didn't click comment on this one:

> I'm able to get multiple authors synced, but with the standard guest author approach from CAP I'm unable to get the guest authors distributed.

I copied the behavior from [the existing code that handles guest authors](https://github.com/Automattic/newspack-network/blob/ca549f3a9ff9110e3507685206f1056ca88880ea/includes/distributor-customizations/class-authorship-filters.php#L47) It would be great to also have it work in admin – at least in the posts listing 🤔 It works for me. They only show on the front end, so might that have been it? Let me know if the front end is not showing them for you.

miguelpeixe:
> They only show on the front end, so might that have been it? Let me know if the front end is not showing them for you.

Ah! My apologies, that's indeed the case. I was expecting it to sync them exactly. Confirmed it's working as intended:

image

naxoc:
Thank you Miguel – I love the idea of filtering the whole post_data part instead. I'll add that tomorrow 👍

naxoc:
OK, @miguelpeixe I added the payload filtering. That is a great idea – it makes everything much more readable and extendable. Ready for re-review :)

tests/unit-tests/content-distribution/util.php
```diff
@@ -22,15 +22,18 @@ function get_sample_payload( $origin = '', $destination = '' ) {
...
 	}
+
 	return [
 		'site_url'         => $origin,
 		'post_id'          => 1,
 		'post_url'         => $origin . '/2021/01/slug',
 		'network_post_id'  => '1234567890abcdef1234567890abcdef',
 		'sites'            => [ $destination ],
 		'status_on_create' => 'draft',
+		'multiple_authors' => [],
```
miguelpeixe:
```suggestion
```

miguelpeixe (approved):
Thanks, @naxoc! Looks good and tests well 💯 

One small ✂️ and we're good to go!

tests/unit-tests/content-distribution/util.php
naxoc: Oh, good catch. Thanks!

PR #198, opened 2025-01-21
fix(content-distribution): prevent consecutive dispatches

miguelpeixe:
p1737411007755239-slack-C055SRT7WN4

When saving a post via Gutenberg, there's a high chance that at least 2 requests will be made to save the post. (see https://github.com/WordPress/gutenberg/issues/15094)

This will cause consecutive dispatches to distribute the post, potentially with the same payload. This PR prevents dispatches with the same payload from getting through.

### Testing

2. Make changes in the editor and save
3. Confirm only 1 dispatch is made and the change is distributed
4. Add a new custom field via the "Custom Fields" advanced section and save
5. Confirm that also only 1 dispatch is made and the change is distributed
6. Make changes to the content, add a new custom field and save
7. Confirm that 2 dispatches are made and the change is distributed (the first will be post content and the second with post meta)

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -21,6 +21,9 @@ use WP_Post;
  * Main class for content distribution
  */
 class Content_Distribution {
+
+	const PAYLOAD_HASH_META = '_newspack_network_payload_hash';
+
 	/**
 	 * Queued network post updates.
 	 *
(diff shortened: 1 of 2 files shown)


naxoc (approved):
Nice! I tested this and it absolutely works. I love the idea with the hash. 👍 

ok

PR #199, opened 2025-01-21
fix(content-distribution): handling multiple post meta

miguelpeixe:
The handling of a post meta with multiple entries doesn't work well. It fails to identify existing entries, adding new values every sync.

This PR changes the strategy by detecting whether the entire group has changed and replacing it. This is also to support multiple entries with equal value, which is allowed.

The protected method also got its name updated to not confuse with core's `update_post_meta`.

### Testing

2. Using the Custom Fields UI, add multiple post meta values sharing the same key and save
3. Confirm they are added to the destination
4. Edit the post again, delete some items, and add others with equal value
5. Confirm the distributed post meta matches exactly

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -21,6 +21,9 @@ use WP_Post;
  * Main class for content distribution
  */
 class Content_Distribution {
+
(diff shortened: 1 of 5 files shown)


naxoc (approved):
Yep! And good call on updating the function name. 👍

PR #200, opened 2025-01-22
feat(content-distribution): remove "Quick Edit" from linked posts

naxoc:
Because quick edit let's the user do stuff we don't want on linked posts, this just removes it.

## How to test
On a node site check that all linked posts don't show "quick edit" on hover on the posts listing. Networked posts that are not linked should still have it. 

Posts not networked should also have it (note that distributor also hides it, so not those posts).

### DIFF
--- a/includes/content-distribution/class-admin.php
+++ b/includes/content-distribution/class-admin.php
@@ -7,11 +7,12 @@
 
 namespace Newspack_Network\Content_Distribution;
 
-use Newspack_Network\Site_Role;
-use Newspack_Network\Hub\Nodes;
-use Newspack_Network\Hub\Node;
 use Newspack\Data_Events;
 use Newspack_Network\Admin as Network_Admin;
+use Newspack_Network\Content_Distribution as Content_Distribution_Class;
+use Newspack_Network\Hub\Node;
+use Newspack_Network\Hub\Nodes;
+use Newspack_Network\Site_Role;
 
 /**
  * Content Distribution Admin Page Class.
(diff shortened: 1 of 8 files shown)


includes/content-distribution/class-admin.php
```diff
@@ -7,11 +7,12 @@
 
 namespace Newspack_Network\Content_Distribution;
 
-use Newspack_Network\Site_Role;
-use Newspack_Network\Hub\Nodes;
-use Newspack_Network\Hub\Node;
 use Newspack\Data_Events;
 use Newspack_Network\Admin as Network_Admin;
+use Newspack_Network\Content_Distribution as Content_Distribution_Class;
```
miguelpeixe, line 12: We've not been using an alias for this in other files, should we start doing so? I thin it'd be good to standardize this.

miguelpeixe (approved):
Works! Left a minor non-blocking comment.

includes/content-distribution/class-admin.php
naxoc: I agree. It's super confusing naming without it, so that's why I aliased it, but I'm up for either. Do you have a preference?

miguelpeixe: I'm ok with the alias, do you mind doing the same for other files?

naxoc: Not at all. Will do and push to this one 👍

naxoc:
I aliased all the `Content_Distribution` class use statements in f3117181f54ed6732445d3a0f38ddf49cad0ad66

miguelpeixe (approved):
Thank you!

PR #203, opened 2025-01-23
feat(testing): Add co-authors-plus to unit testing

naxoc:
This adds CAP so it's always enabled in unit tests. It will need to be specifically disabled in tests if testing something that does not want CAP.

## Todo
- [ ] Add a test that uses CAP 

## How to test
It's for tests! I guess run all tests


naxoc:
Closing this one. Not needed.

PR #204, opened 2025-01-28
chore(content-distribution): better readability for ignored attributes

miguelpeixe:
Update the wording of reserved post meta and taxonomies to **ignored**. It also updates docblocks so the behavior of post meta sync is more clear.

### Testing

There should be no functional changes. Unit tests should pass and distribution should work as expected.


naxoc (approved):
This makes a ton of sense. Thanks! 🙏

PR #205, opened 2025-01-28
feat(content-distribution): partial payload

miguelpeixe:
Implements support for distribution of partial payloads, given a list of post data keys to include in the payload.

The `Content_Distribution::distribute_post_partial( $post, $post_data_keys )` can be called directly to trigger a partial update, but it should be most commonly used in the context of `distribute_queued_posts()`, which will decide on PHP shutdown whether to execute a partial or full distribution of a given post, based on the contents of the `$queued_distributions` var, populated via the `queue_post_distribution( $post_id, $post_data_key );` method.

E.g., if a post is saved and a post meta is updated in the same execution, the following will be called:

```
Content_Distribution::queue_post_distribution( <post_id> ); // via wp_after_insert_post hook
Content_Distribution::queue_post_distribution( <post_id>, 'post_meta' ); // via updated_post_meta hook
```

The above will trigger a single full post update on shutdown.

If only a post meta is updated on the execution:

```
Content_Distribution::queue_post_distribution( <post_id>, 'post_meta' ); // via updated_post_meta hook
```

A single partial update will run on shutdown.

This PR also improves the post meta coverage by hooking into `added_post_meta` and `deleted_post_meta`.

### Testing

2. Update a post meta via CLI `wp post meta update {post_id} test_meta test_value`
3. Confirm in the hub's event log that the payload's post data only includes `post_meta`, `date_gmt`, and `modified_gmt`
4. Delete the post meta via CLI `wp post meta delete {post_id} test_meta`
5. Confirm in the hub's event log table that the payload includes the `post_meta` and `test_meta` is not there
6. In the dashboard, edit the post and save
7. Confirm in the hub's event log that the full payload is sent

### DIFF
--- a/includes/content-distribution/class-incoming-post.php
+++ b/includes/content-distribution/class-incoming-post.php
@@ -87,9 +87,6 @@ class Incoming_Post {
 			throw new \InvalidArgumentException( esc_html( $error->get_error_message() ) );
 		}
 
-		$this->payload         = $payload;
-		$this->network_post_id = $payload['network_post_id'];
-
 		if ( ! $post ) {
 			$post = $this->query_post();
 		}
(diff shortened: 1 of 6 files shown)


includes/content-distribution/class-incoming-post.php
```diff
@@ -173,6 +182,31 @@ protected function get_post_payload() {
...
+	 */
+	protected function get_payload_from_partial( $payload ) {
+		if ( ! $this->ID ) {
+			return new WP_Error( 'missing_post', __( 'Partial payload requires an existing post.', 'newspack-network' ) );
+		}
+
+		$current_payload       = $this->get_post_payload();
+		$current_payload_error = self::get_payload_error( $current_payload );
+		if ( is_wp_error( $current_payload_error ) ) {
+			return $current_payload_error->get_error_message();
```
naxoc: The error should be returned here I think –not a string.

naxoc (approved):
Tested this and it works great. Thanks!

I only added a small nit on return type.

includes/content-distribution/class-incoming-post.php
miguelpeixe: Good catch! Fixed in ac91081baaf6d1312071dbdddb85e8644e7a78c7

PR #207, opened 2025-02-03
fix(content-distribution): use payload hash on partial updates

miguelpeixe:
When partial updates were implemented (#205) it didn't integrate with the payload hash strategy (#198), so every partial update is dispatched even if it's already the most recent distributed data.

This PR implements the payload hash check to partial updates.

It also adds the `_pingme` and `_encloseme` meta to the list of ignored meta keys. These 2 keys are WP internals to manage pings and enclosures via wp-cron and should not be part of the payload.

### Testing

1. On a distributed post, make a content change and save and confirm only 1 dispatch is logged in the hub
2. Make a CAP author change and save and also confirm only 1 dispatch is logged
3. Make a content change, a CAP author change, save, and confirm 2 dispatches are logged
4. Without making any changes, hit Save, and confirm nothing gets dispatched

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -313,6 +313,8 @@ class Content_Distribution {
 			'_edit_last',
 			'_thumbnail_id',
 			'_yoast_wpseo_primary_category',
+			'_pingme',
+			'_encloseme',
 		];
 
 		/**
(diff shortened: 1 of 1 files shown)


naxoc (approved):
Tests well and looks good!

For testing CAP save on it's own I had to save with `wp.data.dispatch('core/editor').savePost()` – I think there is a bug in the "Authors" side panel that doesn't let me save if all I change is authors. Totally unrelated to this though! Good to go 👍

miguelpeixe:
Thanks, @naxoc!

PR #211, opened 2025-02-06
fix(content-distribution): link to post on incoming

naxoc:
Small fix: the "View origin post" on incoming posts was linking to the original site, not post.

See pamTN9-aEn-p2#comment-13129

## How to test
- On an incoming post, click the button and verify that you are taken to the post and not the homepage

### DIFF
--- a/includes/content-distribution/class-editor.php
+++ b/includes/content-distribution/class-editor.php
@@ -109,7 +109,7 @@ class Editor {
 			'newspack-network-incoming-post',
 			'newspack_network_incoming_post',
 			[
-				'originalUrl' => $incoming->get_original_site_url(),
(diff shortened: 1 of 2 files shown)


miguelpeixe (changes requested):
`originalUrl` is also used when displaying the `'Originally distributed from %s.'` and `'Distributed from %s.'` notices.

We should expose `originalSiteUrl` and `originalPostUrl`.

naxoc:
> We should expose originalSiteUrl and originalPostUrl.

Woops! Yes. Absolutely. I added that :)

miguelpeixe (approved):
Thanks for revising!

PR #213, opened 2025-02-12
fix(content-distribution): sync non-publish posts

miguelpeixe:
1208510338230630-as-1209335539861544

The logic to prevent distributing a post when it's not published is at the instantiation level for the Outgoing_Post object, which creates issues when an already published and distributed post moves to draft. We need to be able to instantiate the object, since it has already been distributed.

This causes issues when attempting to sync a distributed post when it moves to non-publish statuses.

This PR removes that restriction from the constructor and applies to the distribution API callback.

### Testing

1. While on trunk, distribute a post with status on create set to `publish`
2. Move the post to draft and confirm that the change does not propagate
3. Checkout this branch, repeat steps 1 and 2 and confirm the status changes are synced

### DIFF
--- a/includes/content-distribution/class-api.php
+++ b/includes/content-distribution/class-api.php
@@ -118,6 +118,15 @@ class API {
 			return new WP_Error( 'newspack_network_content_distribution_error', $e->getMessage(), [ 'status' => 400 ] );
 		}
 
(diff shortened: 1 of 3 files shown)


naxoc (approved):
Yup. Tested this and it looks good!

PR #219, opened 2025-02-21
fix: build on release

miguelpeixe:
Because of content distribution, we just recently started using js assets in the plugin. Releases are missing the `dist/` directory because we're not building the assets on release. This PR fixes that.

See the CI job for reference: https://github.com/Automattic/newspack-scripts/blob/trunk/src/jobs/release.yml#L17


dkoo (approved):

miguelpeixe:
Confirmed that the new alpha (https://github.com/Automattic/newspack-network/releases/tag/v2.7.0-alpha.2) package includes the `dist/` directory.

PR #222, opened 2025-02-26
fix(content-distribution): ignore unchanged values when shortcircuiting meta

miguelpeixe:
Do not shortcircuit the distribution meta if it's unchanged or empty values. Gutenberg may send an empty array for `newspack_network_distributed_sites`, which will reach the shortcircuit strategy and prevent the entire post save.

### Testing

I can't consistently reproduce the editor sending that empty array. Double-checking the new test assertions should be enough.

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -166,6 +166,12 @@ class Content_Distribution {
 			return $check;
 		}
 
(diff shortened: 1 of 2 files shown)


naxoc (approved):
Makes sense!

PR #225, opened 2025-02-28
feat(content-distribution): remove distribution on incoming post deletion

miguelpeixe:
Sends the linked post deletion to the distribution origin so it's removed from the post's distribution configuration. This prevents the post from getting recreated on updates, assuming that the intention of that deletion is to no longer distribute the post to that site.

Our short-circuit strategy prevents a distribution URL from getting removed on post meta updates. To preserve that layer of protection, a `newspack_network_remove_distribution` meta is being introduced for the `remove_distribution( $site_url )` method. This meta contains the URL that's getting intentionally removed to be checked against in the short-circuit callback.

### Testing

2. In the node, permanently delete the post
3. Wait for the `network_incoming_post_deleted` event to reach the origin site
4. Edit the post in the origin site and confirm it's no longer distributing to that node

### DIFF
--- a/includes/content-distribution/class-outgoing-post.php
+++ b/includes/content-distribution/class-outgoing-post.php
@@ -123,10 +123,6 @@ class Outgoing_Post {
 	 * @return true|WP_Error True on success, WP_Error on failure.
 	 */
 	public static function validate_distribution( $urls ) {
-		if ( empty( $urls ) ) {
-			return new WP_Error( 'no_site_urls', __( 'No site URLs provided.', 'newspack-network' ) );
-		}
-
 		if ( in_array( get_bloginfo( 'url' ), $urls, true ) ) {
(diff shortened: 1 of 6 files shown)


includes/content-distribution/class-outgoing-post.php
```diff
@@ -154,6 +150,10 @@ public static function validate_distribution( $urls ) {
 	 * @return array|WP_Error Config array on success, WP_Error on failure.
 	 */
 	public function set_distribution( $site_urls ) {
+		if ( empty( $site_urls ) ) {
+			return new WP_Error( 'no_site_urls', __( 'No site URLs provided.', 'newspack-network' ) );
+		}
+
```
miguelpeixe, line 156: The empty URL validation moved from `validate_distribution()` to here because removing a distribution can result in an empty array and should be valid.

naxoc (approved):
Yep. Tested this and it looks good 👍

PR #227, opened 2025-03-05
feat(event-log): migrate `data` db column to `longtext`

miguelpeixe:
The introduction of content distribution requires the event log data column to hold contents much larger than 64kb. The column type should be migrated to `longtext`. This PR also introduces the use of [dbDelta()](https://developer.wordpress.org/reference/functions/dbdelta/) to manage table changes.

## Testing

1. While on `trunk`, publish a post ensuring it exceeds 64kb in text (you can generate text using https://lipsum.com/)
2. Attempt to distribute that post to a node and confirm the event log is never created and the post doesn't reach the destination
3. Checkout this branch, publish a new post exceeding 64kb and distribute to a node
4. Confirm the event log item is created and the post reaches the destination
5. Run `wp option get newspack_db_version_event_log` and confirm the value is `2`
6. Run `wp db columns wp_newspack_hub_event_log` and confirm `data` is `longtext`

### DIFF
--- a/includes/hub/database/class-event-log.php
+++ b/includes/hub/database/class-event-log.php
@@ -48,9 +48,9 @@ class Event_Log {
 	 */
 	protected static function maybe_update_db() {
 		$db_version = absint( get_option( self::get_current_option_name(), 0 ) );
+		update_option( self::get_current_option_name(), self::DB_VERSION );
 		if ( $db_version < self::DB_VERSION ) {
 			self::update_db();
-			update_option( self::get_current_option_name(), self::DB_VERSION );
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


includes/hub/database/class-event-log.php
```diff
@@ -47,32 +47,39 @@ protected static function get_current_option_name() {
 	 * @return void
 	 */
 	protected static function maybe_update_db() {
-		$db_version = get_option( self::get_current_option_name(), 0 );
+		$db_version = absint( get_option( self::get_current_option_name(), 0 ) );
 		update_option( self::get_current_option_name(), self::DB_VERSION );
```
leogermani, line 51: I think you want to also delete this line

miguelpeixe: We still want the db version to be tracked so it runs `dbDelta()` if necessary.

miguelpeixe: Nevermind, I now realize it's a duplicated line

leogermani:
On my site I ended up on an infinite loop trying to update the DB table when visiting the Event Log. 

Fixed by moving the `update_option` call up and outside the `if ( $db_version...` as it was before.

Also tested deleting the table and checking if it would work well on the first table creation. It did :)

leogermani (approved):

PR #228, opened 2025-03-17
feat(content-distribution): support page post type

miguelpeixe:
Support page distribution.

### Testing

1. Draft a new page and confirm you see the image plugin icon
2. Publish the page and distribute it to a node
3. In the destination confirm the page has been created 
4. Click to edit the page and confirm it's linked and the editor locked
5. Confirm all the labels refer to "page" instead of "post"
6. Unlink and confirm the editor becomes available and you're able to make changes

PR #230, opened 2025-03-17
fix(content-distribution): pages table column

miguelpeixe:
#228 failed to recognize that the `manage_posts_columns` covers the columns for posts and custom post types, but not pages.

This PR adds the Content Distribution column to the Pages table.

### Testing

Visit the Pages admin page and confirm the content distribution column is present in the table.


naxoc (approved):
Works! 👍

PR #234, opened 2025-03-25
fix(woocommerce): tweaks for membership and subscription events

adekbadek:
### All Submissions:

Two small tweaks:

1. When a membership is updated via the network, the membership note is not specifying which status was propagated (only the fact of status change). This PR adds the status name to the note.
2. Fixes a deprecation warning when using `last_payment_date`:

```
PHP Deprecated:  Function WC_Subscription::get_date was called with an argument that is <strong>deprecated</strong> since version 2.2.0! The &quot;last_payment&quot; date type parameter has been deprecated due to ambiguity (it actually returns the date created for the last order) and to align date types with improvements to date APIs in WooCommerce 3.0, specifically the introduction of a new &quot;date_paid&quot; API. Use &quot;last_order_date_created&quot; or &quot;last_order_date_paid&quot;
```

1. On `trunk`, renew a subscription*. Observe a deprecation warning logged. 
4. On this branch, observe no deprecation warning. 
5. Update a status of a network-propagated membership, observe the membership note on the "non-original" site (the one where the membership is only mirrored) contains the status. 

\* "process renewal" in "Order actions":

image

### Other information:

### DIFF
--- a/includes/woocommerce/class-events.php
+++ b/includes/woocommerce/class-events.php
@@ -97,7 +97,7 @@ class Events {
 		$result['start_date'] = $item->get_date( 'start_date' );
 		$result['trial_end_date'] = $item->get_date( 'trial_end_date' );
 		$result['next_payment_date'] = $item->get_date( 'next_payment_date' );
-		$result['last_payment_date'] = $item->get_date( 'last_payment_date' );
+		$result['last_payment_date'] = $item->get_date( 'last_order_date_paid' );
 		$result['end_date'] = $item->get_date( 'end_date' );
 		$result['products'] = [];
 
--- a/includes/woocommerce/class-events.php
+++ b/includes/woocommerce/class-events.php
@@ -97,7 +97,7 @@ class Events {
 		$result['start_date'] = $item->get_date( 'start_date' );
 		$result['trial_end_date'] = $item->get_date( 'trial_end_date' );
 		$result['next_payment_date'] = $item->get_date( 'next_payment_date' );
-		$result['last_payment_date'] = $item->get_date( 'last_order_date_paid' );
+		$result['last_payment_date'] = $item->get_date( 'last_order_date_created' );
 		$result['end_date'] = $item->get_date( 'end_date' );
 		$result['products'] = [];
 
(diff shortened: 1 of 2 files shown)


includes/woocommerce/class-events.php
```diff
@@ -97,7 +97,7 @@ public static function subscription_changed( $item_id, $status_from, $status_to,
 		$result['start_date'] = $item->get_date( 'start_date' );
 		$result['trial_end_date'] = $item->get_date( 'trial_end_date' );
 		$result['next_payment_date'] = $item->get_date( 'next_payment_date' );
-		$result['last_payment_date'] = $item->get_date( 'last_payment_date' );
+		$result['last_payment_date'] = $item->get_date( 'last_order_date_paid' );
```
leogermani: This will radically change the value of this column.

`last_order_date_paid` will not consider the last successful order. So if a subscription's last order is a failed order, this will return zero.

But, this value is actually used in the Subscriptions table in the consolidated Subscripions list under the label of "Last Order Date":

image

This table is meant to mimic the native Woo Subscriptions table. And from what I see, what's returned there is `last_order_date_created` -> https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/admin/class-wcs-admin-post-types.php#L687

So I think we should use the same here.

I also tested the behavior on a Subscription that had some recent failed orders, and I confirmed that what we see in the table is not `last_order_date_paid`, but `created`.

Of course this creates a mismatch between the table label and the slug of the column... but that's also the case for Woo Subscriptions core so... 

I had this on top of my mind because of the recent changes in https://github.com/Automattic/newspack-plugin/pull/3842/files ... We need to be careful with those date functions

leogermani (changes requested):

includes/woocommerce/class-events.php
adekbadek: Thank you for pointing this out, and the thorough explanation! Updated in a571993c7f4107f331c8bbe1c5a9c4745bed9973

leogermani (approved):

PR #236, opened 2025-04-14
feat: adjust capabilities

adekbadek:
### All Submissions:

Adjusts caps, so the unavailable menu options are hidden for non-admins.

_Dependent on https://github.com/Automattic/newspack-plugin/pull/3920_

1. On `trunk`, 
2. Create a new author user and log in as them
4. Observe the Network → Subscriptions and → Orders admin menus are displayed

image

4. Switch to this branch, observe these items are not displayed
5. Log in a the site administrator, confirm you can access these and other Network page via the admin menu

### Other information:

### DIFF
--- a/includes/hub/database/class-orders.php
+++ b/includes/hub/database/class-orders.php
@@ -81,7 +81,9 @@ class Orders {
 			'can_export'       => false,
 			'capability_type'  => 'post',
 			'capabilities'     => [
-				'create_posts' => 'not_a_real_capability', // Set to a fake capability to remove "add new" button.
+				'create_posts'      => 'not_a_real_capability', // Set to a fake capability to remove "add new" button.
+				'edit_posts'        => \Newspack_Network\Admin::REQUIRED_CAPABILITY,
+				'edit_others_posts' => \Newspack_Network\Admin::REQUIRED_CAPABILITY,
 			],
 			'show_in_rest'     => false,
 			'delete_with_user' => false,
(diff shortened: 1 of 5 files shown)


includes/hub/database/class-orders.php
```diff
@@ -81,7 +81,9 @@ public static function register_post_type() {
 			'can_export'       => false,
 			'capability_type'  => 'post',
 			'capabilities'     => [
-				'create_posts' => 'not_a_real_capability', // Set to a fake capability to remove "add new" button.
+				'create_posts'      => 'not_a_real_capability', // Set to a fake capability to remove "add new" button.
+				'edit_posts'        => \Newspack_Network\Admin::REQUIRED_CAPABILITY,
```
naxoc: Nit: it's the same namespace, so \Network_Network\ is not needed

naxoc (changes requested):
I tested this locally. It does indeed make the menu item go away, but it is also gone for my admin user. Should the capability be added/registered somewhere for it to work?

includes/hub/database/class-orders.php
adekbadek: Thanks! Fixed in [aca3096](https://github.com/Automattic/newspack-network/pull/236/commits/aca30966af04f3fd3a4c4cf69554666f39bc6766)

adekbadek:
> I tested this locally. It does indeed make the menu item go away, but it is also gone for my admin user. Should the capability be added/registered somewhere for it to work?

Ah, that's true. This will depend on https://github.com/Automattic/newspack-plugin/pull/3920, then, when it's merged.

PR #238, opened 2025-05-05
feat: add option to not create new terms on content distrib NPPD-134

leogermani:
Add a new check to define taxonomies that should be distributed but for which we don't want to create new terms in the target site if they don't exist.

This is useful for the integration with Multibranded sites plugin, in which we don't want to propagate all the brands from all the sites all other sites. We just want to keep the posts in the same brand if they exist in more than one site.

Read the code and see the tests.

If you really want to test it:

* Activate Multibranded site in 2 sites in the network
* Create 2 brands in site A
* Creates 2 brand in site B, being one of the existing brands in site A
* On site A, mark the post in both brands and distribute it to site B
* Confirm that on site B the post is assigned to the existing brand, but the second brand is not created (on trunk, the second brand will be created)

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -416,6 +416,24 @@ class Content_Distribution {
 	}
 
 	/**
+	 * Returns a list of taxonomies that should be distributed only if the terms already exist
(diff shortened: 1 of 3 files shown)


miguelpeixe (approved):
Clean solution and tests well!

PR #240, opened 2025-05-23
feat: sync post modified date on post ditribution

leogermani:
On Content Distrbution, when processing an incoming post, make sure that the post modified date is kept in sync with the original post.

## Testing

1. Distribute a post
2. Check the post in the destination and confirm it has the exact same post_modified date from the original
3. Make changes to the original post and save
4. Confirm that the new modified date is synced

### DIFF
--- a/includes/content-distribution/class-incoming-post.php
+++ b/includes/content-distribution/class-incoming-post.php
@@ -691,6 +691,9 @@ class Incoming_Post {
 
 			// Handle taxonomy terms.
 			$this->update_taxonomy_terms();
+
+			// Handle the post modified date.
+			$this->update_post_modified_date();
 		}
 
 		update_post_meta( $this->ID, self::PAYLOAD_META, $this->payload );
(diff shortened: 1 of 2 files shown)


naxoc:
I tested this and it works. Code looks OK. I'm not entirely sure I see the use case though. Is it in case the original post's modified date gets changed to the past? Otherwise I think the truth is reflected by the updated modified that will naturally trigger when the node updates with new data?

leogermani:
> I tested this and it works. Code looks OK. I'm not entirely sure I see the use case though. Is it in case the original post's modified date gets changed to the past? Otherwise I think the truth is reflected by the updated modified that will naturally trigger when the node updates with new data?

Thanks for the review.

One thing you must know: There`s an option in the theme that displays the post modified date in the front end, for the readers, if the date differs from the published date ( ex: "Published Oct 4, Last updated Oct 15)

When I publish a post and immediately distribute it, the publish and modified date are identical on the original post, bu they differ by a couple of minutes in the target post. Not a big deal from a data standpoint, but this ends up being shown in the article page, as if the post has had an update after it was published.. it looks wrong.

Also, if I published a post yesterday, and distributed it today, the post in the target site will have an Updated date displayed to the readers that says it was modified today, which is not true

naxoc (approved):
Ah, thanks! Those are good reasons. In that case this is good to go! 👍

PR #241, opened 2025-05-23
feat: add support to yoast primary cat in content distribution

leogermani:
On content distribution, distribute Yoast's primary category if present

## Testing

1. Test Content distribution without Yoast and confirm it still works
2. enable Yoast
3. Create a post with multiple categories and set a primary category
4. Distribute the post
5. Confirm that in the destination the correct primary category is set, even if the ID is not the same in both sites
6. Update the post and change the primary category
7. confirm the update also works

### DIFF
--- a/includes/content-distribution/class-yoast-primary-cat.php
+++ b/includes/content-distribution/class-yoast-primary-cat.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Newspack Network Distributor Yoast Primary Category on Content Distribution
+ *
+ * @package newspack-network
+ */
+
+namespace Newspack_Network\Content_Distribution;
+
+/**
+ * Class to handle Yoast's primary category metadata distribution
+ */
+class Yoast_Primary_Cat {
+
+	/**
+	 * The meta name for the primary category slug.
+	 *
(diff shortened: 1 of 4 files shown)


includes/content-distribution/class-yoast-primary-cat.php
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * Filter the outgoing post meta.
+	 *
+	 * @param array   $meta The post meta.
+	 * @param WP_Post $post The post object.
+	 * @return array
+	 */
+	public static function filter_outgoing_post( $meta, $post ) {
+
+		if ( ! class_exists( 'WPSEO_Primary_Term' ) ) {
```
naxoc, line 39: Because the classes are called with the backslash prefix elsewehere, it might be good to consistently do that and use it in the cheks to see if it exists too.

naxoc (approved):
Tested and looks good. I only had a small nit.

includes/content-distribution/class-yoast-primary-cat.php
leogermani: class_exists always look from the root...

PR #243, opened 2025-05-27
feat(content-distribution): pull posts and Story Budget integration [NPPD-127]

miguelpeixe:
Implements pull for content distribution, to be first implemented as a Story Budget action:

## Single

image

---

image

## Bulk

image

---

image

## Next steps

The integration class implements the Sites fields, but not yet the distribution integration on field updates. In the next iteration, the field will become read-only as we'll rely only on the distribution data. This depends on allowing the distribution of non-published posts.

### Status on publish

It's not yet implemented, but the "status on create" functionality will change to "status on publish". We'll soon allow a distribution to happen before publication. In that case, the selected status will be applied upon publication and not upon post creation in the destination.

## Testing

1. Checkout this branch and https://github.com/Automattic/newspack-story-budget/pull/89
2. Make sure you have both installed in all nodes in your network
3. In one node, go through the steps to connect to a remote site (see https://github.com/Automattic/newspack-story-budget/pull/84)
4. Confirm you see the "Pull" action as seen above for single entries and bulk when selecting stories
5. Pull a published story and confirm the modal opens and the story is pulled after selecting the "Status on publish"
6. Visit your posts dashboard and confirm the story is pulled with the correct status
7. Select multiple stories, pull and confirm they are all pulled with the selected status

---
  - https://app.asana.com/0/0/1209924286461760

### DIFF
--- a/includes/class-initializer.php
+++ b/includes/class-initializer.php
@@ -51,6 +51,7 @@ class Initializer {
 		Distributor_Customizations::init();
 		Esp_Metadata_Sync::init();
 		Content_Distribution::init();
+		Story_Budget::init();
 
 		Synchronize_All::init();
 		Data_Backfill::init();
--- a/includes/class-story-budget.php
+++ b/includes/class-story-budget.php
@@ -113,7 +113,7 @@ class Story_Budget {
 			}
 		}
 
-		$field = Fields::get_field( 'network_sites' );
+		$field = \Newspack_Story_Budget\Fields::get_field( 'network_sites' );
 		return \get_post_meta( $post_id, $field->get_post_meta_name(), false );
 	}
 
--- a/includes/class-story-budget.php
+++ b/includes/class-story-budget.php
@@ -164,15 +164,16 @@ class Story_Budget {
 			return $metadata;
 		}
 
+		$metadata['can_pull']  = true;
 		$metadata['is_pulled'] = false;
-
-		$outgoing_post = Content_Distribution::get_distributed_post( $story_id );
-		if ( ! $outgoing_post ) {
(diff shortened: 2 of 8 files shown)


leogermani:
This is looking great!

Here's what I've found:

On a node, the Hub does not show up as an option to connect to. It should

The Pull icon is not rendering for me (neither in the row nor in the bulk actions):

image

Now the rest I think it's not yet covered by this PR

* The post doesn't land in any budget, so it never shows up in SB
* Need to figure out how to do it. I think we've mentioned using the same "Add new" dialog, so the fields that show up there could be edited right away. But they need to be able to at least select the target budget
 * Not sure how Bulk pull would work in that context... (and I think it's fine if we don't have bulk pulling)
* Need to add an exception to the meta fields that are synced. SB meta fields should not be synced (just the first time)
*

miguelpeixe:
> The Pull icon is not rendering for me (neither in the row nor in the bulk actions):

Forgot to mention, you need to run an `npm ci` because of the updated `@wordpress/icons` package.

leogermani:
> On a node, the Hub does not show up as an option to connect to. It should

@miguelpeixe do you intend to address this on this PR?

miguelpeixe:
I missed that comment, I was waiting for your review. I'll address that and ping you for another review.

The other points should be addressed separately.

miguelpeixe:
Fixed in d385b8a200d42944e0fdda6cca45ae415cefb044. Because we don't store the hub name in the node, I'm using a pretty URL as label (removing `https://www.` and trailing slash)

leogermani (approved):

PR #244, opened 2025-05-29
Distribution tax hierarchy

leogermani:
This PR makes Content Distribution account for taxonomy hierarchy

## Testing

* Test distributing posts and assigning categories nested under other categories
* Test it with both categories that exist and don't exist in the destination site
* The whole hierarchy should be recreated in the target site
* Also confirm that the Yoast's primary category is correctly assigned when hierarchy is involved

### DIFF
--- a/includes/content-distribution/class-taxonomy-terms.php
+++ b/includes/content-distribution/class-taxonomy-terms.php
@@ -77,8 +77,6 @@ class Taxonomy_Terms {
 	 * @return array|WP_Error The term IDs on success, WP_Error on failure.
 	 */
 	public static function get_or_create_term_ids( $terms, $taxonomy ) {
-		$ignored_taxonomies = Content_Distribution_Class::get_ignored_taxonomies();
-
 		$term_ids = [];
 
 		foreach ( $terms as $term_data ) {
--- a/includes/content-distribution/class-taxonomy-terms.php
+++ b/includes/content-distribution/class-taxonomy-terms.php
@@ -17,13 +17,49 @@ class Taxonomy_Terms {
 	const SEPARATOR = '|--|';
 
 	/**
+	 * Get taxonomies that should not be distributed.
+	 *
+	 * @return string[] The ignored taxonomies.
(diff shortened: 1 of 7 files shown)


includes/content-distribution/class-taxonomy-terms.php
```diff
@@ -0,0 +1,149 @@
...
+	const SEPARATOR = '|--|';
+
+	/**
+	 * Get post taxonomy terms for distribution.
+	 *
+	 * @param \WP_Post $post The post object.
+	 * @return array The taxonomy term data.
+	 */
+	public static function get_post_taxonomy_terms( \WP_Post $post ) {
+		$ignored_taxonomies = Content_Distribution_Class::get_ignored_taxonomies();
```
miguelpeixe: non-blocking: We could move this here now that we have a dedicated class for taxonomies.

includes/content-distribution/class-taxonomy-terms.php
```diff
@@ -0,0 +1,149 @@
...
+					'parent'     => $parent,
+					'fields'     => 'ids',
+				]
+			);
+
+			if ( empty( $found_terms ) ) {
+
+				// If the taxonomy is in the list of taxonomies that should only
+				// have terms that already exist, skip the term creation.
+				if ( in_array( $taxonomy, Content_Distribution_Class::get_existing_terms_only_taxonomies(), true ) ) {
```
miguelpeixe: Same here, this class is a fitting home for `get_existing_terms_only_taxonomies()`

miguelpeixe (approved):
Works great! A couple of non-blocking suggestions below.

includes/content-distribution/class-taxonomy-terms.php
leogermani: Great feedback. Addressed in https://github.com/Automattic/newspack-network/pull/244/commits/e0674fdde1ae78a50cd5c5e49e2cedb23796ede2

PR #247, opened 2025-06-11
feat(story-budget): manage field syncing and props

miguelpeixe:
Manage synced and unsynced fields for distributed posts and additional field props control:

 - For incoming posts, synced fields and the `network_sites` field should not be editable
 - All the other fields (unsynced) should be ignored on distribution

### Testing

1. Checkout this branch and https://github.com/Automattic/newspack-story-budget/pull/98
2. Edit a few story fields on a site and distribute that story to multiple sites
3. Visit the Story Budget app in the, edit the "Sites" field and confirm the options are disabled (can't undistribute)
4. Go to one of the destination sites and set a budget for the incoming post
5. View the linked story in the Story Budget app and enable "Edit mode"
6. Confirm the "Status" and "Sites" fields are not editable
7. Confirm that the "Sites" fields render the origin and other destinations (not the current site)
8. Confirm that the field values did not sync
9. Back to the origin site, edit the "Status" field, and wait for distribution
10. Go to the destination site and confirm that the "Status" field updated

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -53,6 +53,7 @@ class Content_Distribution {
 		add_filter( 'newspack_webhooks_request_priority', [ __CLASS__, 'webhooks_request_priority' ], 10, 2 );
 		add_filter( 'update_post_metadata', [ __CLASS__, 'maybe_short_circuit_distributed_meta' ], 10, 4 );
 		add_action( 'wp_after_insert_post', [ __CLASS__, 'handle_post_updated' ] );
+		add_action( 'set_object_terms', [ __CLASS__, 'handle_post_updated' ] );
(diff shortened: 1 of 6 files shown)


leogermani (changes requested):
I could not confirm that the status change on the origin post is updated on the distributed post. The payload of the network event doesn't have anything about the status (we're talking about the StoryBudget status, right? not the wp post status)

Also, not sure if this is subject for another PR, but I think ideally we could copy the values of the meda fields in the first sync, but then leave them editable and not update them again in future syncs

miguelpeixe:
Not sure why I thought the Status field was getting synced. It was not handled at all because it's a taxonomy. Added support in 917feb3685a385f18dc68e0f4b1e1879354eae5d.

The Story Name is also synced now.

leogermani (approved):

PR #248, opened 2025-06-12
feat: add network subscriptions to my account NPPM-350

leogermani:
Add network Subscriptions to My Account

If a reader purchases a subscription in one site, there's no way for them to move the subscription management to another site in the network if they wish to.

This PR displays the subscriptions the user has purchased in other sites in the network, with a handy link for them to manage them there.

image

## Implementation notes

At first, I thought it would make sense to simply add it to the native "Subscriptions" screen from Woo, but there are some challenges there:

1. This page won't even show up if the user doesn't have a subscription in that site
2. It also won't show up if they user has only one subscription in that site. (A "My subscription" item will be displayed instead)

So it's not trivial to plug our list to those screens.

Also, if the user has a subscription in other sites, it's likely that they don't have any on this site.

## Testing

1. Setup a network
2. Buy a subscription in a site
3. Wait for the events to propagate
4. On the other site, log in as the reader who made the purchase and go to My Account
5. You should see the new Network Subscriptions page

### DIFF
--- a/includes/class-initializer.php
+++ b/includes/class-initializer.php
@@ -58,7 +58,7 @@ class Initializer {
 		Membership_Dedupe::init();
 
 		Woocommerce\Events::init();
-
+		Woocommerce_Subscriptions\My_Account::init();
 		Woocommerce_Memberships\Admin::init();
 		Woocommerce_Memberships\Events::init();
 		Woocommerce_Memberships\Subscriptions_Integration::init();
--- a/includes/woocommerce-subscriptions/class-my-account.php
+++ b/includes/woocommerce-subscriptions/class-my-account.php
@@ -63,7 +63,16 @@ class My_Account {
 	 * @return array The menu items.
 	 */
 	public static function add_menu_item( $items ) {
+
+		$user = wp_get_current_user();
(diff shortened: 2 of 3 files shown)


thomasguillot:
A couple of thoughts:
* “Network Subscriptions” might be too abstract for users. They may not realise the site they’re on is part of a network, or know what that means. Could we rename it to “Other Subscriptions” instead?
* Can we move this menu item below “My Subscription”?
* The table looks a bit tight around the button — could we have 16px of padding above and below for better spacing?

leogermani:
Thanks for the suggestions @thomasguillot . Latest commits:

* Add some CSS to the table with some spacing
* Change the menu positionn
* Conditionally change the menu item label. If the user doesn't have any other subscriptions in this site, the menu item is labeled simply "Subscriptions", otherwise, it's labeled "Other Subscriptions"

## No other subscription

image

## Just one subscription in the site

Captura de tela de 2025-06-16 15-47-30

## More than one subscription on the site

Captura de tela de 2025-06-16 15-48-08

thomasguillot:
Thank you for the changes, @leogermani. This looks much better now.

adekbadek (approved):

PR #250, opened 2025-06-13
feat(content-distribution): distribute non-publish content

miguelpeixe:
Support the distribution of non-published content.

The "status on create" option has been refactored to "status on publish". When the post is distributed as a draft or any non-publish status, the selected "status on publish" is stored as meta, and the distributed post will sync the origin status until publication. Once published, the destination will apply the stored "status on publish" and remove the meta.

This only happens once. If the post gets unpublished, it will not get automatically republished via sync. It has to be republished manually.

The phrasing when selecting the post status was changed when distributing unpublished posts:

| Draft | Publish |
| --- | --- |
| image | image |

### Testing

1. Confirm new unit tests are accurate
2. Start a post draft and distribute it to a node with the "Status on publish" as `publish`
3. Confirm the post reaches the destination as a draft
4. Change the origin post status to `pending` and save
5. Confirm the post updates to `pending`
6. Publish the post in the origin and confirm the post is also published in the node
7. Move the origin post back to draft and confirm the post is set to draft in the destination
8. Publish again and confirm the post is **not** published in the node
9. Create a new draft and distribute with the "Status on publish" as `pending`
10. Confirm the post reaches the destination as a draft
11. Publish the post in the origin and confirm the post is `pending` in the node

### DIFF
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -53,6 +53,7 @@ class Content_Distribution {
 		add_filter( 'newspack_webhooks_request_priority', [ __CLASS__, 'webhooks_request_priority' ], 10, 2 );
 		add_filter( 'update_post_metadata', [ __CLASS__, 'maybe_short_circuit_distributed_meta' ], 10, 4 );
 		add_action( 'wp_after_insert_post', [ __CLASS__, 'handle_post_updated' ] );
+		add_action( 'set_object_terms', [ __CLASS__, 'handle_post_updated' ] );
 		add_action( 'updated_post_meta', [ __CLASS__, 'handle_postmeta_update' ], 10, 3 );
 		add_action( 'added_post_meta', [ __CLASS__, 'handle_postmeta_update' ], 10, 3 );
 		add_action( 'deleted_post_meta', [ __CLASS__, 'handle_postmeta_update' ], 10, 3 );
--- a/includes/class-content-distribution.php
+++ b/includes/class-content-distribution.php
@@ -395,6 +395,7 @@ class Content_Distribution {
 				Incoming_Post::PAYLOAD_META,
 				Incoming_Post::UNLINKED_META,
 				Incoming_Post::ATTACHMENT_META,
+				Incoming_Post::STATUS_ON_PUBLISH_META,
 			]
 		);
 	}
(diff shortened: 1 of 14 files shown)


leogermani (approved):
Works!

There was one time when I saw an error when distributing a post that was still in auto-draft.

When I tried to reproduce the issue, I saw that the UI was showing "Save & Distribute", and worked as expected... Not sure how I got there... I was running the latest built because it was my second test and I saw the new labels there... 

One thing I noticed, not sure if it's a problem:
* Distribute a post with "status on publish" as draft
* publish it on origin
* manually publish it on the target site
* set it to draft in the origin
* It's automatically set to draft in the target site

(here's a proof that I triggered an event with the auto draft)

```
{
    "site_url": "https:\/\/leogermani.sa.ngrok.io",
    "post_id": 261,
    "post_url": "https:\/\/leogermani.sa.ngrok.io\/?p=261",
    "network_post_id": "d83f293f3d63383a47d1884c41570d68",
    "sites": [
        "https:\/\/super.local"
    ],
    "status_on_publish": "draft",
    "post_data": {
        "title": "Auto Draft",
        "author": {
            "type": "wp_user",
            "ID": 1,
            "display_name": "leo",
            "user_email": "wordpress@example.com",
            "user_url": "http:\/\/localhost"
        },
        "post_status": "auto-draft",
        "date_gmt": "0000-00-00 00:00:00",
        "modified_gmt": "0000-00-00 00:00:00",
        "slug": "",
        "post_type": "post",
        "raw_content": "",
        "content": "",
        "excerpt": "",
        "comment_status": "open",
        "ping_status": "open",
        "taxonomy": [],
        "thumbnail_url": false,
        "post_meta": {
            "_wp_page_template": [
                ""
            ]
        }
    }
}
```

miguelpeixe:
> There was one time when I saw an error when distributing a post that was still in auto-draft.

Good catch, added backend restriction in 210db67eddb11ad10a7263ae2addd0eed9e2fd76 and a UI tweak in 01343de147974bc468f320c4a06befb6676bddd4. The side panel should render a notice and disable the form for auto-drafts:

image

I'd consider this a very edge case, though. Why would someone try to distribute an auto-draft?

---

> One thing I noticed, not sure if it's a problem:
> 
> - Distribute a post with "status on publish" as draft
> - publish it on origin
> - manually publish it on the target site
> - set it to draft in the origin
> - It's automatically set to draft in the target site

One way to work around this is to store a meta when the post is published via sync, so it only changes the status from 'publish' to something else if the meta exists. This would ensure that manually published posts never get unpublished. I wonder if we want that behavior and if it's intuitive for publishers. Sometimes these adjustments overcomplicate things.

leogermani (approved):

PR #251, opened 2025-06-17
fix(story-budget): read-only sites field

miguelpeixe:
Because the distribution requires more configuration than just selecting the site, the "Sites" field should be read-only for now.

If we allow distribution, as we do with pull, editing the field should enable setting the "Status on publish" and "Budget" options.

## Testing

Confirm the Sites field is no longer editable.


leogermani (commented):
NIT: fix the doc block for the get_sites_field_value method that still says it will get the value from post meta

leogermani (approved):

PR #258, opened 2025-07-08
Network memberships integrity check

adekbadek:
### All Submissions:

Adds a WP CLI command to verify the network memberships integrity.

NPPM-386

1. Introduce some discrepancies by e.g. changing the status of a network-synced membership directly in the DB (this way it won't be picked up as an event)
2. Run `wp newspack-network integrity-check --verbose` and observe the discrepancies are listed

### Other information:

* [x] Have you written new tests for your changes, as applicable?

### DIFF
--- a/includes/class-integrity-check-utils.php
+++ b/includes/class-integrity-check-utils.php
@@ -1,0 +1,190 @@
+<?php
+/**
+ * Integrity Check utility functions.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network;
+
+use Newspack_Network\Woocommerce_Memberships\Admin as Memberships_Admin;
+
+/**
+ * Shared utility functions for integrity check operations.
+ */
+class Integrity_Check_Utils {
+
+	/**
+	 * Get all membership data
+	 *
+	 * @param int|null $max_records Maximum number of records to return (for testing).
+	 * @return array Array of (email, status, network_id) data
+	 */
+	public static function get_membership_data( $max_records = null ) {
+		if ( ! class_exists( 'WC_Memberships_User_Membership' ) ) {
+			return [];
+		}
+
+		global $wpdb;
+
+		// phpcs:disable WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+		$query = "
+			SELECT 
+				u.user_email,
+				p.post_status as status,
+				pm_network.meta_value as network_id
+			FROM {$wpdb->posts} p
+			INNER JOIN {$wpdb->users} u ON p.post_author = u.ID
+			INNER JOIN {$wpdb->postmeta} pm_network ON p.post_parent = pm_network.post_id AND pm_network.meta_key = %s
+			INNER JOIN (
+				SELECT 
+					p2.post_author,
+					pm2.meta_value,
+					MAX(p2.post_date) as max_date
+				FROM {$wpdb->posts} p2
+				INNER JOIN {$wpdb->postmeta} pm2 ON p2.post_parent = pm2.post_id AND pm2.meta_key = %s
+				WHERE p2.post_type = 'wc_user_membership'
+				AND pm2.meta_value IS NOT NULL
+				AND pm2.meta_value != ''
+				GROUP BY p2.post_author, pm2.meta_value
+			) latest ON p.post_author = latest.post_author 
+				AND pm_network.meta_value = latest.meta_value 
+				AND p.post_date = latest.max_date
+			WHERE p.post_type = 'wc_user_membership'
+			AND pm_network.meta_value IS NOT NULL
+			AND pm_network.meta_value != ''
+			ORDER BY LOWER(u.user_email) ASC
+		";
+		// phpcs:enable WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+
+		if ( $max_records ) {
+			$query .= $wpdb->prepare( ' LIMIT %d', $max_records );
+		}
+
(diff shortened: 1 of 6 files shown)


includes/cli/class-integrity-check.php
```diff
@@ -0,0 +1,629 @@
...
+	 * @return void
+	 */
+	public static function integrity_check( $args, $assoc_args ) { // phpcs:ignore Generic.NamingConventions.ConstructorName.OldStyle
+		$verbose = isset( $assoc_args['verbose'] ) ? true : false;
+		$fix_discrepancies = isset( $assoc_args['fix-discrepancies'] ) ? true : false;
+		$chunk_size = isset( $assoc_args['chunk-size'] ) ? intval( $assoc_args['chunk-size'] ) : 1000;
+		$max_records = isset( $assoc_args['max'] ) ? intval( $assoc_args['max'] ) : null;
+
+		if ( ! Site_Role::is_hub() ) {
+			WP_CLI::error( 'This command can only be run on a Hub site.' );
```
chickenn00dle: Nonblocking: Can we only register this command for hub sites?

includes/cli/class-integrity-check.php
```diff
@@ -0,0 +1,629 @@
...
+					}
+				}
+				$table_data[] = $discrepancy;
+			}
+
+			// Display as table using WP-CLI's table formatter.
+			WP_CLI\Utils\format_items( 'table', $table_data, $node_columns );
+		}
+
+		if ( $fix_discrepancies ) {
```
chickenn00dle: Nonblocking nit: Can we remove this if we're not implementing yet?

includes/cli/class-integrity-check.php
```diff
@@ -0,0 +1,629 @@
...
+
+	/**
+	 * Get chunk data from a node via REST API
+	 *
+	 * @param \Newspack_Network\Node\Node $node The node to query.
+	 * @param int                         $offset The offset for the chunk.
+	 * @param int                         $limit The limit for the chunk.
+	 * @return array The chunk data from the node
+	 */
+	private static function get_node_chunk_data( $node, $offset, $limit ) {
```
chickenn00dle: Seems to be a lot of repeated code for fetching via the REST API. Maybe its worth creating a general API fetch method to clean this up a bit?

includes/cli/class-integrity-check.php
```diff
@@ -0,0 +1,629 @@
...
+	 * ## OPTIONS
+	 *
+	 * [--verbose]
+	 * : Output verbose information during the check.
+	 *
+	 * [--fix-discrepancies]
+	 * : Attempt to fix discrepancies found during the check.
+	 *
+	 * [--chunk-size=<size>]
+	 * : Maximum number of memberships to compare in each chunk (default: 1000).
```
chickenn00dle: This might be expected, but adjusting this option sometimes leads to different results. The lower the number the less discrepencies are found. For example, here is the result with default chunk size

Image

And with a chunk size of 3:

Image

This might be the result of comparing less data at any given time, but I wonder how useful this flag is if it makes the results inconsistent.

includes/class-integrity-check-utils.php
```diff
@@ -0,0 +1,190 @@
...
+	 */
+	public static function get_membership_data_range( $start_email, $end_email, $max_records = null ) {
+		if ( ! class_exists( 'WC_Memberships_User_Membership' ) ) {
+			return [];
+		}
+
+		global $wpdb;
+
+		// phpcs:disable WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
+		$query = "
```
chickenn00dle: Non-blocking: Not sure if its worth it, but this query and the one above look to be almost identical aside from 2 lines. Maybe its worth moving this into its own helper method so we aren't duplicating so much?

includes/cli/class-integrity-check.php
```diff
@@ -0,0 +1,629 @@
...
+		$response = wp_remote_get(
+			$endpoint,
+			[
+				'headers' => $node->get_authorization_headers( 'integrity-check' ),
+				'timeout' => 60, // phpcs:ignore WordPressVIPMinimum.Performance.RemoteRequestTimeout.timeout_timeout
+			]
+		);
+
+		if ( is_wp_error( $response ) || 200 !== wp_remote_retrieve_response_code( $response ) ) {
+			WP_CLI::error( sprintf( 'Failed to get membership data from node: %s', $node->get_url() ) );
```
chickenn00dle, line 182: Just noting that `WP_CLI::error` will halt execution. Not sure if this is what we want, or if the intention is to just report the error and continue with other nodes if available.

chickenn00dle (commented):
Looks good for the most part @adekbadek!

I left a handful of comments, most of which are non-blocking. But I did run into an issue with inconsistent results based on chunk-size.

includes/cli/class-integrity-check.php
adekbadek: Good question! I'd leave it as it as, opting for "full integrity check". If one node can't give us data, then the whole check is not complete.

includes/class-integrity-check-utils.php
adekbadek: Good call, refactored in recent changes

includes/cli/class-integrity-check.php
adekbadek: Good call, refactored in recent changes

includes/cli/class-integrity-check.php
adekbadek: Good call, removed in recent changes

includes/cli/class-integrity-check.php
adekbadek: Good call, updated in recent changes

includes/cli/class-integrity-check.php
adekbadek: Thanks for catching that! Fixed now

chickenn00dle:
Just noting that everything is working and looks good except for the chunk-size behavior which is still inconsistent. I'll take over this PR and see about getting this fixed (or just remove the param for the time being).

miguelpeixe (approved):

PR #262, opened 2025-07-18
Feat/propagate hub name

leogermani:
This PR does two (3) things :) 

1. It propagates the Hub's name to all Nodes, 
2. Makes the Hub name be used in the Top admin bar menu, where it lists the sites and their bookmarks

(and 3. Refactors the registration of the menu bar to reduce code duplication and adds a Utils\Sites method that is useful for integrations to easily get the list of access not having to worry about whether we are in the hub or the node)

## How to test

1. Load this branch and navigate both on the Hub and Nodes and confirm the admin bar menu still works
2. Confirm the Hub is still called "Hub"
3. Run the backfill on the Hub CLI -> `wp newspack-network data-backfill network_hub_name_updated` (dry run) and `wp newspack-network data-backfill network_hub_name_updated --verbose --live`.
4. Confirm the new event Hub Name Updated added to the Event log
5. Pull the events in a Node
6. Confirm the Hub's name now show up in the admin bar menu in the Node
7. Back to the Hub, go to Settings > General and change the site's name
8. Confirm the event is created in the event log
9. Check the methods available in the Utils\Sites class and test them in WP Shell

### DIFF
--- a/includes/utils/class-sites.php
+++ b/includes/utils/class-sites.php
@@ -36,7 +36,7 @@ class Sites {
 	 *
 	 * @return array Array of arrays with name and url properties.
 	 */
-	public static function get_all_sites_but_current() {
+	public static function get_all_sites_without_current() {
 		$sites = self::get_all_sites();
 		return array_filter(
 			$sites,
--- a/includes/utils/class-sites.php
+++ b/includes/utils/class-sites.php
@@ -1,0 +1,148 @@
+<?php
+/**
+ * Newspack Network Sites methods.
+ *
+ * Helper methods to get sites lists and info in an easy way.
+ *
+ * @package Newspack
+ */
+
+namespace Newspack_Network\Utils;
+
+use Newspack_Network\Site_Role;
+use Newspack_Network\Hub\Nodes;
+use Newspack_Network\Hub\Node as Hub_Node;
+use Newspack_Network\Node\Settings as Node_Settings;
+
+/**
+ * Class to watch the user for updates and trigger events
+ */
+class Sites {
+
+	/**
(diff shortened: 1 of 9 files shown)


includes/utils/class-sites.php
```diff
@@ -0,0 +1,151 @@
...
+	public static function get_all_sites() {
+		return array_merge( [ self::get_hub() ], self::get_nodes() );
+	}
+
+	/**
+	 * Get all sites but the current one
+	 *
+	 * @return array Array of arrays with name and url properties.
+	 */
+	public static function get_all_sites_but_current() {
```
adekbadek: Nit: 

```suggestion
	public static function get_all_sites_without_current() {
```

IMO that's more readable

adekbadek (changes requested):
In my testing, in the backfill the names were missing (first line):

```
Backfilling data for action network_hub_name_updated, from  to .

👉 Hub name updated to The Local Tribune.

Success: Processed 1 network_hub_name_updated events, ignored 0 as duplicates and skipped 0.
```

includes/utils/class-sites.php
leogermani: no strong preference: applied in https://github.com/Automattic/newspack-network/pull/262/commits/a742a4492431ecf6201a5f7dd34b6fb0a7f9e8bd

leogermani:
> Backfilling data for action network_hub_name_updated, from  to .

The missing values here are an issue with the backfiller itself, when you don't determine a time range. Outside the scope of this PR

adekbadek (approved):

PR #264, opened 2025-07-30
fix: avoid post status collision

leogermani:
When these post types were first created, we should have added more specific slugs to the custom statuses, because they can conflict with core statuses, as post statuses apply to all post types.

This is in fact happening with "pending", making the core's pending status public.

We could simply remove the "pending" status registration and things would work just fine, as this is already registered by core. But the best thing to do is to add a very specific prefix to avoid these statuses to conflict with anything else in the future.

## Testing

Make sure you have some orders and subscriptions in your Hub's database.

Checkout this branch and make sure things just keep working as usual. The existing items are listed as before, and new items are created and updated as usual.

### DIFF
--- a/includes/hub/admin/css/woo-cpts.css
+++ b/includes/hub/admin/css/woo-cpts.css
@@ -21,51 +21,51 @@
     text-overflow: ellipsis;
 }
 
-.order-status.status-np-net-completed {
+.order-status.status-npn-completed {
     background: #c8d7e1;
     color: #2e4453;
 }
 
-.order-status.status-np-net-on-hold {
+.order-status.status-npn-on-hold {
     background: #f8dda7;
     color: #94660c;
 }
 
-.order-status.status-np-net-failed {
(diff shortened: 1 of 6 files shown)


leogermani:
There's a limit of 20 characters in the post_status column in the DB. `np-net-pending-cancel` is exceeding it. Need to shorten this prefix

leogermani:
Fixed in https://github.com/Automattic/newspack-network/pull/264/commits/c89ded3e7f7ff8c7909d5a8058eb73866f034c9b

In case you have already executed the DB migration and what to test it again, here's how to reset it:

on SQL: `update wp_posts set post_status = SUBSTRING( post_status, 8, 50 ) where post_status like 'np-net-%';`

Then
```
n wp option delete np_hub_orders_db_version
n wp option delete np_hub_subscriptions_db_version
n wp cache flush
```

miguelpeixe (approved):
Works!

PR #291, opened 2026-01-30
fix(sync): cast role to array for backwards compatibility

wil-gerken:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-block-theme/blob/trunk/.github/CONTRIBUTING.md)?

Casts `$data->role` to an array in `class-user-manually-synced.php` to maintain backwards compatibility with pre-#187 events in the event log.

[PR #187](https://github.com/Automattic/newspack-network/pull/187) changed the `role` field from a string to an array on the sending side (`class-user-manual-sync.php`), but the receiving side (`class-user-manually-synced.php`) wasn't made backwards-compatible with events already in the event log.

Nodes that haven't processed past the old-format events get permanently stuck, as `array_diff()` is erroring when it receives a string instead of an array. 

Since this error occurs right before `set_last_processed_id()` in `process_pulled_data()`, the last processed ID never advances, and the node re-pulls the same failing batch indefinitely.

You can see the mixed types in the `network_manual_sync_user` events here:

7218 "role":"subscriber" ← string (pre-#187)
36531 "role":"customer" ← string
...
122559 "role":["subscriber"] ← array (post-#187)
146730 "role":["contributor"] ← array

The (array) cast handles both the old string format and the new array format. With a string value like "subscriber", it becomes ["subscriber"]. With an array value like ["subscriber", "customer"], it passes through unchanged. Both cases produce a valid array for array_diff() 

### Other information:

* [ ] Have you successfully ran tests with your changes locally?

I'm not able to test this full loop locally, but I have tested that the (array) cast will work with both a string and an array.

### DIFF
--- a/includes/incoming-events/class-user-manually-synced.php
+++ b/includes/incoming-events/class-user-manually-synced.php
@@ -80,7 +80,7 @@ class User_Manually_Synced extends Abstract_Incoming_Event {
 
 		// Update user role if changed.
 		$user_current_roles = $user->roles;
-		$user_new_roles     = $data->role ?? [];
+		$user_new_roles     = (array) ( $data->role ?? [] );
 		$remove_roles       = array_diff( $user_current_roles, $user_new_roles );
 		$add_roles          = array_diff( $user_new_roles, $user_current_roles );
 


leogermani (commented):
@wil-gerken Let's make this a hotfix:

* Branch it off from the `release` branch
* prefix the branch name with `hotfix/`
* submit a new PR based to `release`

wil-gerken:
Thank you and done! PR #292

chickenn00dle:
Can we close this one now @wil-gerken?

wil-gerken:
@chickenn00dle, oh sorry about that. Yes! I didn't realize this was still open. Thanks for the heads up.

PR #307, opened 2026-03-25
feat(integrity-check): add --fix flag to reconcile membership discrepancies

adekbadek:
### All Submissions:

* [x] Have you followed the [Newspack Contributing guideline](https://github.com/Automattic/newspack-block-theme/blob/trunk/.github/CONTRIBUTING.md)?

Adds a `--fix` flag to the existing `wp newspack-network integrity-check` command. When passed, the command goes beyond reporting discrepancies and actively reconciles them by dispatching membership update events.

Addresses NPPM-386.

**How it works:**
1. The integrity check runs as usual, finding discrepancies between hub and nodes.
2. With `--fix`, each discrepancy is classified and an action is determined:
   - **Missing on node** (hub has it, node doesn't): dispatches a membership update event so the node creates the membership on its next sync. Skipped (`skip_no_plan`) if the node doesn't have the plan.
   - **Missing on hub** (node has it, hub doesn't): if the node membership is backed by a **local subscription** (an authoritative source), dispatches an event so the hub creates the membership (`pull_to_hub`). If it has no subscription it may be a stale managed mirror, so it is flagged `skip_no_subscription` for manual review rather than pulled -- pulling is dispatched as a hub-originated event that every other node then pulls, so a stale record would otherwise be resurrected network-wide.
   - **Transfer** (node has it under old email, hub has it under new email): detected via remote_id cross-referencing. The node's managed membership `_remote_id` is matched to the hub's `membership_id`. Dispatches a transfer event with `previous_email` so the node reassigns the membership using the transfer handling from #306.
   - **Status mismatch**: the side with a **local subscription** attached is authoritative -- a subscription (any status) is the source of truth for membership status. Hub has a subscription and node doesn't → `push_to_node`; node has one and hub doesn't → `pull_to_hub`; both or neither have a subscription → hub wins by default (`push_to_node`). Membership modification timestamps are used only to stamp dispatched events (for event-log dedup and ordering), not to decide the winner.
3. A table of planned actions is displayed. Reconciliation then asks for confirmation (`--yes` to skip in non-interactive runs) before dispatching events with a progress bar.

**Sync lag safety:**
- Before dispatching, `--fix` queries each node's sync status via a new `/integrity-check/sync-status` REST endpoint.
- If a node has unprocessed events in its queue, the command aborts and suggests running `wp newspack-network sync-all` on the node first -- discrepancies may resolve after syncing without creating new events. The node's own events are excluded from this comparison (mirroring the `/pull` endpoint) so a node's own emitted events are not counted as phantom lag.
- If a node's sync status cannot be confirmed (endpoint unreachable), reconciliation for that node is skipped so it is not mutated against a state we could not verify.
- Use `--force` to skip the sync lag check and reconcile nodes with an unconfirmed status.
- If managed-memberships data cannot be fetched from a node, reconciliation for that node is skipped entirely.

**Supporting changes:**
- `Integrity_Check_Utils` includes `post_modified_gmt`, `membership_id`, and `has_subscription` in membership data queries (hashing is unchanged). Cross-site timestamps use `post_modified_gmt` (UTC).
- A new `/integrity-check/managed-memberships` REST endpoint on nodes returns managed membership details (`_remote_id`, `post_modified_gmt`, `has_subscription`), enabling transfer detection and subscription-based authority.
- A new `/integrity-check/sync-status` REST endpoint on nodes returns the `last_processed_id` for sync lag detection and the node's available plan network IDs.

1. Set up a Hub + Node network with WooCommerce Memberships and at least one membership plan with a `_newspack_network_id`.
2. Run `wp newspack-network integrity-check --verbose` on the hub to confirm existing discrepancies are reported as before.
3. Run `wp newspack-network integrity-check --fix --verbose` on the hub. Confirm the reconciliation prompt appears (and that `--yes` skips it).
4. Verify the output shows a reconciliation section with an action table per node (columns: email, network_id, type, hub_status, node_status, action) and a summary of dispatched vs skipped actions.
5. Check the hub's event log for newly created `newspack_network_woo_membership_updated` events, and verify nodes pick them up on their next pull cycle.
6. **Transfer:** transfer a membership from User A to User B on the hub, run `--fix`, and verify the action is `push_transfer` with the correct `previous_email`. Confirm the node reassigns the membership to User B.
7. **Status mismatch (subscription authority):** create a status mismatch where the node membership has a local subscription but the hub's does not; run `--fix` and verify the action is `pull_to_hub`. Then create the inverse (hub has the subscription) and verify `push_to_node`. With neither side holding a subscription, verify the hub wins (`push_to_node`).
8. **Missing on hub gating:** create a network-managed membership on a node with no local subscription that the hub lacks; run `--fix` and verify the action is `skip_no_subscription` (not pulled). Repeat with a subscription-backed node membership and verify `pull_to_hub`.
9. **Sync lag / unknown status:** ensure a node has unprocessed events, run `--fix`, and verify it aborts with a suggestion to sync first; run with `--force` to confirm it proceeds. Make a node's sync-status endpoint unreachable and verify that node is skipped without `--force`.

### Other information:

* [x] Have you written new tests for your changes, as applicable?


includes/cli/class-integrity-check.php
```diff
@@ -494,4 +555,190 @@ private static function get_node_range_data( $node, $start_email, $end_email, $m
...
+							$email,
+							$network_id,
+							$hub_status,
+							$hub_modified,
+							$node_status,
+							$node_modified
+						)
+					);
+				}
+				$action = 'skip';
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

includes/cli/class-integrity-check.php
```diff
@@ -494,4 +555,190 @@ private static function get_node_range_data( $node, $start_email, $end_email, $m
...
+			'user_id'         => 0,
+			'plan_network_id' => $hub_item['network_id'],
+			'membership_id'   => $hub_item['membership_id'],
+			'new_status'      => str_replace( 'wcm-', '', $hub_item['status'] ),
+		];
+
+		$event = new \Newspack_Network\Incoming_Events\Woocommerce_Membership_Updated(
+			get_bloginfo( 'url' ),
+			$event_data,
+			time()
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

includes/node/class-integrity-check-endpoints.php
```diff
@@ -148,6 +162,55 @@ public static function handle_memberships_request( $request ) {
...
+				FROM {$wpdb->posts} p
+				INNER JOIN {$wpdb->users} u ON p.post_author = u.ID
+				INNER JOIN {$wpdb->postmeta} pm_managed ON p.ID = pm_managed.post_id AND pm_managed.meta_key = %s
+				LEFT JOIN {$wpdb->postmeta} pm_remote ON p.ID = pm_remote.post_id AND pm_remote.meta_key = %s
+				LEFT JOIN {$wpdb->postmeta} pm_site ON p.ID = pm_site.post_id AND pm_site.meta_key = %s
+				LEFT JOIN {$wpdb->postmeta} pm_network ON p.post_parent = pm_network.post_id AND pm_network.meta_key = %s
+				WHERE p.post_type = 'wc_user_membership'",
+				'_managed_by_newspack_network',
+				'_remote_id',
+				'_remote_site_url',
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

tests/unit-tests/test-reconcile-memberships.php
```diff
@@ -0,0 +1,331 @@
...
+/**
+ * Class TestReconcileMemberships
+ *
+ * @package Newspack_Network
+ */
+
+use Newspack_Network\CLI\Integrity_Check;
+
+/**
+ * Test the Reconcile_Memberships::classify_discrepancies method.
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

includes/node/class-integrity-check-endpoints.php
```diff
@@ -148,6 +162,55 @@ public static function handle_memberships_request( $request ) {
...
+					pm_remote.meta_value as remote_id,
+					pm_site.meta_value as remote_site_url,
+					pm_network.meta_value as network_id
+				FROM {$wpdb->posts} p
+				INNER JOIN {$wpdb->users} u ON p.post_author = u.ID
+				INNER JOIN {$wpdb->postmeta} pm_managed ON p.ID = pm_managed.post_id AND pm_managed.meta_key = %s
+				LEFT JOIN {$wpdb->postmeta} pm_remote ON p.ID = pm_remote.post_id AND pm_remote.meta_key = %s
+				LEFT JOIN {$wpdb->postmeta} pm_site ON p.ID = pm_site.post_id AND pm_site.meta_key = %s
+				LEFT JOIN {$wpdb->postmeta} pm_network ON p.post_parent = pm_network.post_id AND pm_network.meta_key = %s
+				WHERE p.post_type = 'wc_user_membership'",
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

includes/node/class-integrity-check-endpoints.php
```diff
@@ -148,6 +162,55 @@ public static function handle_memberships_request( $request ) {
 		);
 	}
 
+	/**
+	 * Handles the managed memberships request.
+	 *
+	 * Returns all network-managed memberships with their remote_id and remote_site_url.
+	 *
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

includes/cli/class-integrity-check.php
```diff
@@ -157,6 +162,62 @@ public static function integrity_check( $args, $assoc_args ) { // phpcs:ignore G
...
+
+				// Get managed memberships from node for timestamp comparison.
+				$node_managed = self::get_node_managed_memberships( $node );
+
+				// Get full node membership data.
+				$node_data = self::get_node_membership_data( $node );
+
+				// Classify discrepancies.
+				$classified = self::classify_discrepancies( $hub_lookup, $node_data, $node_managed );
+
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

includes/class-integrity-check-utils.php
```diff
@@ -26,10 +26,12 @@ private static function build_membership_query( $start_email = null, $end_email
 
 		// phpcs:disable WordPressVIPMinimum.Variables.RestrictedVariables.user_meta__wpdb__users
 		$query = "
-			SELECT 
+			SELECT
 				u.user_email,
 				p.post_status as status,
-				pm_network.meta_value as network_id
+				pm_network.meta_value as network_id,
+				p.post_modified,
```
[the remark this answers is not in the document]
adekbadek: Addressed in cda0e25.

includes/cli/class-integrity-check.php
```diff
@@ -157,6 +174,144 @@ public static function integrity_check( $args, $assoc_args ) { // phpcs:ignore G
...
+				// Count actionable items for progress.
+				$actionable = array_filter(
+					$classified,
+					function( $item ) {
+						return in_array( $item['action'], [ 'push_to_node', 'push_transfer', 'pull_to_hub' ], true );
+					}
+				);
+				$node_total = count( $actionable );
+				$node_done  = 0;
+
```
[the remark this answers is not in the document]
adekbadek: Addressed in 29552b0.

tests/unit-tests/test-reconcile-memberships.php
```diff
@@ -0,0 +1,429 @@
...
+
+	/**
+	 * When the node has a membership the hub doesn't, it is classified as
+	 * missing_on_hub with action skip.
+	 */
+	public function test_missing_on_hub_results_in_skip_action() {
+		$classify_discrepancies_method = $this->get_classify_discrepancies_method();
+
+		$hub_lookup = [];
+
```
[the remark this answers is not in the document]
adekbadek: Addressed in 29552b0.

includes/cli/class-integrity-check.php
```diff
@@ -492,6 +658,416 @@ private static function get_node_range_hash( $node, $start_email, $end_email, $m
...
+				// Node has it, hub does not – pull from node to hub.
+				$discrepancies[] = [
+					'email'       => $email,
+					'network_id'  => $network_id,
+					'type'        => 'missing_on_hub',
+					'hub_status'  => '',
+					'node_status' => $node_status,
+					'action'      => 'pull_to_hub',
+					'node_data'   => $node_item,
+				];
```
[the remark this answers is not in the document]
adekbadek: Addressed in 29552b0 (updated PR description to reflect bidirectional reconciliation).

includes/cli/class-integrity-check.php
```diff
@@ -440,8 +602,8 @@ private static function get_node_range_request( $node, $endpoint_type, $start_em
 		$endpoint = sprintf( '%s/wp-json/newspack-network/v1/integrity-check/%s', $node->get_url(), $endpoint_type );
 
 		$query_args = [
-			'start' => strtolower( $start_email ),
-			'end'   => strtolower( $end_email ),
+			'start' => rawurlencode( strtolower( $start_email ) ),
+			'end'   => rawurlencode( strtolower( $end_email ) ),
```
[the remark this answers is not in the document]
adekbadek: Addressed in 29552b0.

includes/cli/class-integrity-check.php
```diff
@@ -157,6 +174,196 @@ public static function integrity_check( $args, $assoc_args ) { // phpcs:ignore G
...
+		}
+
+		if ( $fix ) {
+			WP_CLI::line( '' );
+
+			// Query sync status and plan availability from all nodes.
+			$node_plan_ids  = []; // Keyed by node URL.
+			$nodes_behind   = [];
+
+			foreach ( $discrepancies as $node ) {
```
jason10lee, line 246: **Blocker:** A node whose sync-status check fails is still reconciled, bypassing the lag guard — If `get_node_sync_status()` returns null (endpoint unreachable / timeout), the node is `continue`d out of the lag-check loop (`:241-245`), so it never lands in `$nodes_behind` and never triggers the abort — yet the reconcile loop at `:281` iterates all discrepancies and dispatches to that same node anyway. Net effect: a node we *could not confirm* is caught-up gets its memberships mutated, silently skipping the sync-lag safety the feature advertises — **even without `--force`**. **Risk if not addressed:** the operator believes the lag check protected the run, but a node that was actually behind (merely unreachable for the status probe) is reconciled against stale data. Skip reconciliation (or require `--force`) for nodes whose sync status is unknown.

**Verified live:** with the node's `sync-status` endpoint forced to return 500, `--fix` (no `--force`) printed `Warning: Could not check sync status` and then reconciled the node anyway (`Dispatched: 1`) — the lag guard was silently skipped.

includes/cli/class-integrity-check.php
```diff
@@ -492,6 +710,437 @@ private static function get_node_range_hash( $node, $start_email, $end_email, $m
...
+			// A subscription (any status) is the source of truth for membership status.
+			$hub_has_sub  = ! empty( $hub_item['has_subscription'] );
+			$node_has_sub = ! empty( $node_item['has_subscription'] );
+
+			if ( $hub_has_sub && ! $node_has_sub ) {
+				$action = 'push_to_node';
+			} elseif ( ! $hub_has_sub && $node_has_sub ) {
+				// Node has a subscription, hub does not – node is authoritative.
+				$action = 'pull_to_hub';
+			} else {
```
jason10lee, line 900: **Blocker:** `--fix` can overwrite a newer node status with the hub's older one (no recency check) — The PR description says status mismatches resolve by *newer wins* via `post_modified_gmt`, and that *the hub is not blindly trusted*. But `classify_discrepancies()` performs **no timestamp comparison** — it picks the authority purely from `has_subscription`, and when both or neither side has a subscription it unconditionally does `push_to_node` (hub wins, line 855). `parse_gmt_timestamp()` is used only to stamp dispatched events for dedup, never to classify. So when a reader holds subscriptions on two sites, or a manual status change happened on the node with no subscription either side, a genuinely newer node-side status is silently overwritten by the hub's older status. The timestamps needed to do this correctly are already collected — reconcile the code with the stated design, or at minimum correct the description.

**Verified locally:** the PR's own `test-reconcile-memberships.php` passes 9/9, and `test_status_mismatch_no_subscriptions_defaults_to_hub` asserts `push_to_node`; no test drives the outcome from `post_modified`. The 'newer wins' policy is neither implemented nor tested.

**Risk if not addressed:** Because the hub-wins default fires across every status-mismatched paid membership in the network at once and the receiving side overwrites without checking which side changed last, a batch of live paid memberships is reset to a stale status with no per-record recovery state captured — the same out-of-order-status failure that caused the original incident, now executed in bulk; undoing it requires reconstructing each membership's correct status by hand.

includes/cli/class-integrity-check.php
```diff
@@ -492,6 +710,437 @@ private static function get_node_range_hash( $node, $start_email, $end_email, $m
...
+	 * Dispatch a membership_updated event from node data to create the membership on the hub.
+	 *
+	 * Used for missing_on_hub discrepancies: the node has a membership that the hub doesn't.
+	 * Creates an event attributed to the node so the hub processes it locally.
+	 *
+	 * @param array  $node_item A node membership record (email, status, network_id, membership_id).
+	 * @param string $node_url  The node's URL (used as the event's originating site).
+	 * @return void
+	 */
+	private static function dispatch_to_hub( $node_item, $node_url ) {
```
jason10lee, line 1147: **Blocker:** `pull_to_hub` propagates one node's membership network-wide with no staleness gate — A membership present on a node but missing on the hub is dispatched via `dispatch_to_hub()` and persisted with `node_id` = the source node. Because `/pull` excludes only the *source* node, **every other node** then pulls the event and creates/updates the membership too — carrying that one node's status, with no check on whether the node's record is stale. `missing_on_hub` is the one discrepancy type with no subscription/recency gating at all. A single node that erroneously still shows a membership active (cancelled elsewhere but not synced) can resurrect that member — and their paid access — across the whole network.

**Verified (mechanism):** a live hub↔node `--fix` run confirmed reconciliation events are dispatched *hub-originated* — `dispatch_to_node()` builds the event with `get_bloginfo('url')` as the source and calls `process_in_hub()`, so it lands in the event log with `node_id=0` and is therefore pulled by **every** node, not just the one with the discrepancy. (For `push_to_node` a broadcast is expected; the risk here is the `pull_to_hub` direction propagating one node's status network-wide.)

**Risk if not addressed:** If a single node still shows a membership active that was actually cancelled or expired elsewhere and never synced, reconciliation treats that node as the source and resurrects the active status across the entire network, re-granting paid access to a reader who should have none — and reversing it means chasing the same membership across every site.

includes/cli/class-integrity-check.php
```diff
@@ -157,6 +174,196 @@ public static function integrity_check( $args, $assoc_args ) { // phpcs:ignore G
...
+						WP_CLI::line( sprintf( '    ... and %d more', count( $items ) - 5 ) );
+					}
+				}
+			}
+			WP_CLI::line( '' );
+			WP_CLI::line( 'These memberships are sources (they have a local subscription) but are incorrectly marked as network-managed.' );
+			WP_CLI::line( 'To fix: remove _managed_by_newspack_network, _remote_id, and _remote_site_url meta from these memberships.' );
+		}
+
+		if ( $fix ) {
```
jason10lee, line 238: **Suggestion:** No confirmation prompt before the destructive `--fix` — `--fix` dispatches network-wide membership create/update/transfer events with no `WP_CLI::confirm()`. The only guard is the sync-lag check, which `--force` bypasses, after which events fire immediately. WP-CLI convention for a destructive, hard-to-reverse operation that mutates membership state across every node is to prompt (`WP_CLI::confirm( …, $assoc_args )`) so a `--yes` is required for non-interactive runs — especially since `--force` already removes the one existing rail.

**Verified:** a live `--fix` run dispatched the membership event immediately, with no prompt.

includes/cli/class-integrity-check.php
```diff
@@ -157,6 +174,196 @@ public static function integrity_check( $args, $assoc_args ) { // phpcs:ignore G
...
+
+			foreach ( $discrepancies as $node ) {
+				$sync_status = self::get_node_sync_status( $node );
+				if ( null === $sync_status ) {
+					continue;
+				}
+				$node_plan_ids[ $node->get_url() ] = $sync_status['plan_network_ids'];
+
+				if ( ! $force ) {
+					$hub_latest_id = self::get_hub_latest_event_id();
```
jason10lee: **Suggestion:** `get_hub_latest_event_id()` recomputed inside the per-node loop — `$hub_latest_id = self::get_hub_latest_event_id();` is recomputed on every iteration of `foreach ( $discrepancies as $node )` though it has no node dependency — each call runs an `Event_Log::get()` query. Hoist it out of the loop (compute once before it) to avoid N redundant event-log queries.

includes/cli/class-integrity-check.php
```diff
@@ -492,6 +710,437 @@ private static function get_node_range_hash( $node, $start_email, $end_email, $m
...
+		return $circulars;
+	}
+
+	/**
+	 * Get the latest pullable event ID from the hub event log,
+	 * avoiding false positives from non-pullable events like order_changed.
+	 *
+	 * @return int The latest event ID, or 0 if the log is empty.
+	 */
+	private static function get_hub_latest_event_id() {
```
jason10lee: **Suggestion:** Sync-lag guard false-positives on any node that has emitted a membership event — `get_hub_latest_event_id()` (line ~1019) returns the max event-log id across `ACTIONS_THAT_NODES_PULL` **without** `excluded_node_id`, so it counts events that originated from the node being checked. But the hub `/pull` endpoint excludes a node's own events, and `Pulling::process_pulled_data()` only advances the cursor to ids it actually received. `newspack_network_woo_membership_updated` is a pullable action nodes themselves emit — so whenever the latest pullable hub event belongs to the node being checked, that node's `last_processed_id` stays permanently below `hub_latest_id`, the lag check reports phantom 'unprocessed events', and `--fix` aborts (`:258`) telling the operator to run `sync-all` (which can't help). It fails *safe* (over-abort), but it affects essentially every active node, so operators will learn to always pass `--force` — removing the only guard in front of the mutation paths. Exclude the node's own events from the comparison (as `/pull` does).

**Verified live:** injecting one node-originated event made `--fix` report `1 unprocessed events (last processed: 1, hub latest: 2)` and abort — while `sync-all` on the node left `last_processed` at 1 (it can't pull its own event), so the advised remedy can't clear the phantom lag.

includes/node/class-integrity-check-endpoints.php
```diff
@@ -148,6 +170,60 @@ public static function handle_memberships_request( $request ) {
...
 
+	/**
+	 * Returns all network-managed memberships.
+	 *
+	 * Each membership entry includes: email, status, network_id, remote_id,
+	 * remote_site_url, post_modified (GMT), and membership_id.
+	 *
+	 * @param \WP_REST_Request $request The REST request object.
+	 */
+	public static function handle_managed_memberships_request( $request ) {
```
jason10lee, line 181: **Suggestion:** `managed-memberships` endpoint returns all memberships unpaginated (OOM risk at scale) — `handle_managed_memberships_request()` and the base membership query (`class-integrity-check-utils.php:24`) return **all** network-managed memberships in a single response with no range/LIMIT — unlike the hash endpoints, which are range-chunked. On a node with a large `wc_user_membership` table the `INNER JOIN (… GROUP BY …)` forces an on-disk temp table plus a large serialized payload. Given this runs remotely on every node during `--fix` and the target networks have 40k+ memberships (the scale that OOM'd the prior experimental audit), range-paginate this the way the hash endpoints already are.

includes/cli/class-integrity-check.php
```diff
@@ -157,6 +174,196 @@ public static function integrity_check( $args, $assoc_args ) { // phpcs:ignore G
...
+					WP_CLI::line( '  No actionable discrepancies.' );
+					continue;
+				}
+
+				// Display action table.
+				$action_columns = [ 'email', 'network_id', 'type', 'hub_status', 'node_status', 'action' ];
+				WP_CLI\Utils\format_items( 'table', $classified, $action_columns );
+
+				// Count actionable items for progress.
+				$actionable = array_filter(
```
jason10lee, line 341: **Nit:** Progress-bar tick count can under-run the total — `$node_total` counts every actionable item, but the dispatch loop skips `tick()` when a `push_to_node` item has no `$hub_lookup` entry (`:333`) or a `pull_to_hub` item has an empty `membership_id` (`:348`), so the bar `finish()`es short of 100%. Cosmetic only.

includes/cli/class-integrity-check.php
```diff
@@ -492,6 +710,437 @@ private static function get_node_range_hash( $node, $start_email, $end_email, $m
...
+
+	/**
+	 * Parse a GMT timestamp string into a Unix timestamp.
+	 *
+	 * Uses explicit UTC timezone to avoid dependence on the server's default timezone.
+	 *
+	 * @param string $date_string A date string in 'Y-m-d H:i:s' format (GMT).
+	 * @return int|false Unix timestamp, or false on parse failure.
+	 */
+	private static function parse_gmt_timestamp( $date_string ) {
```
jason10lee, line 1122: **Nit:** `parse_gmt_timestamp()` doesn't reject WordPress zero-dates — `DateTimeImmutable::createFromFormat('Y-m-d H:i:s', …)` returns a truthy object for `'0000-00-00 00:00:00'` (a value WP can store), yielding a large-negative `getTimestamp()` rather than `false`, so the `time()` fallback never engages and a negative event timestamp is stored. Impact is low (status application ignores the timestamp; dedup stays deterministic), but checking `getLastErrors()` / rejecting non-positive results is more robust.

jason10lee (changes requested):
Clearing up pre-monorepo PRs!  Great idea to build a tool for this, and the bones of it are sound: transfer detection is strongly cross-referenced (`remote_id` ↔ `membership_id`), dispatch is idempotent, and the new endpoints reuse the existing signature auth and prepared statements.  This also was an opportunity to test whether we could get a meaningful local test of networked nodes (turns out Claude's perfectly capable of setting it up--though make sure you've got copies of commercial plugins handy).

Agentic review has flagged three paths that can corrupt access at scale and look like blockers: 
1.  the "newer wins" policy in the description isn't actually implemented — ties default to the hub, so a newer node status can be overwritten with an older hub one (the same out-of-order-status failure NPPM-386 is about);
2. `pull_to_hub` propagates one node's membership network-wide with no staleness check, so a single stale 'active' can re-grant access everywhere; and
3. a node whose sync-status probe fails is reconciled anyway, skipping the lag guard even without `--force`. Alongside those: the lag guard itself false-positives on essentially every active node (so operators will habitually `--force`), there's no `WP_CLI::confirm`, and the `managed-memberships` endpoint is unpaginated — the OOM mode that killed the prior audit. 

As for *non-blocking* suggestions, we recommend implementing the timestamp tiebreaker (the data is already collected), requiring an interactive confirm, writing a recoverable before/after log of every change, and paginating the endpoint. 

There *are* nits too, but they're tiny (zero-date parsing, docblocks, an `explode` limit, the progress bar).

The membership data mutated here is live paid access across a multi-site network and the dispatch is effectively irreversible without per-record manual repair. Before treating `--fix` as merge-ready it should: compare the two sides' modification timestamps to choose the winner (the data is already collected), require an interactive confirmation, and write a recoverable before/after log of every record it changes.

**Considerations affecting code outside the patch**:
- `includes/cli/class-integrity-check.php:374` — **Nit:** Docblock cleanups (return types, wrong Node class, copy-paste) — A few docs are out of sync with the code: `get_node_membership_data` / `get_node_hash` docblocks say `@return array` / `string` but return `null` (`:374`, `:404` → should be `?array` / `?string`); several `@param` docblocks reference `\Newspack_Network\Node\Node` while `Nodes::get_all_nodes()` actually yields `\Newspack_Network\Hub\Node` (`:373` etc.); and `get_local_circular_links()`'s docblock (`:957`) is a copy-paste of `get_hub_latest_event_id()`'s summary.
- `includes/cli/class-integrity-check.php:551` — **Nit:** `explode('::', $key)` missing the 2-arg limit in `compare_chunk_data` — `classify_discrepancies()` uses `explode( '::', $key, 2 )` (`:800`) to stay robust against a `::` inside the value, but `compare_chunk_data` (`:551`) omits the limit. Harmless for current data (emails / network_ids), but use the 2-arg form for parity.

includes/cli/class-integrity-check.php
adekbadek: Addressed in 9cdc901. Nodes whose sync-status probe fails are now recorded and skipped in the reconcile loop (with a warning) unless `--force` is passed, so an unreachable status check no longer lets a node slip past the lag guard and get mutated against an unverified state.

includes/cli/class-integrity-check.php
adekbadek: You're right that the code performs no timestamp comparison — that was an intentional move (c350b2d) to subscription-presence authority, and the PR description was stale. I've corrected the description to match the implementation: the side with a local subscription is authoritative; when both or neither side has one, the hub wins by default; timestamps are used only to stamp dispatched events for event-log dedup/ordering, never to classify. The `missing_on_hub` path is now gated on the node holding a subscription too (see the pull_to_hub thread / 9cdc901), so a subscription-less node status is never treated as the source of truth.

includes/cli/class-integrity-check.php
adekbadek: Addressed in 9cdc901. `missing_on_hub` is now dispatched as `pull_to_hub` only when the node membership is backed by a local subscription (an authoritative source); a subscription-less managed mirror is flagged `skip_no_subscription` for manual review instead of pulled. That closes the "single stale node resurrects a membership network-wide" path, since a hub-originated pull is what every other node then pulls. The status-mismatch `pull_to_hub` branch was already gated on the node holding the subscription. Transfer detection still reclassifies genuine transfers regardless of the new action, and there's a test asserting that.

includes/cli/class-integrity-check.php
adekbadek: Addressed in 9cdc901. Added `WP_CLI::confirm( …, $assoc_args )` immediately before the dispatch loop (after the lag check passes), so interactive runs prompt and non-interactive runs require `--yes`. `--yes` is documented in the command synopsis.

includes/cli/class-integrity-check.php
adekbadek: Good catch on the redundant call, but this is superseded by the sync-lag fix in the neighbouring thread: `get_hub_latest_event_id()` now takes an `excluded_node_id` and is called per-node to exclude that node's own events (matching `/pull`). The value is therefore genuinely node-dependent and can't be hoisted out of the loop — it stays in the loop by necessity now.

includes/cli/class-integrity-check.php
adekbadek: Addressed in 9cdc901. `get_hub_latest_event_id()` now accepts `excluded_node_id` and the lag check passes `$node->get_id()`, so events a node itself emitted are excluded from its latest-event comparison — mirroring the `/pull` endpoint — and no longer report phantom lag that `sync-all` on that node can never clear.

includes/node/class-integrity-check-endpoints.php
adekbadek: I looked into this and don't think the OOM risk holds for this endpoint. `handle_managed_memberships_request()` is a plain join (`posts`↔`users`↔`postmeta`) with no `GROUP BY` and no derived table, so there's no on-disk temp table — the `INNER JOIN (… GROUP BY …)` you're describing is in `build_membership_query()` (the `/memberships` + `/range-data` path), a different query. For the flagged endpoint at 40k rows × ~7 short columns the serialized payload is a few MB, well within PHP's memory limit, and this has been run against production networks at that scale without OOMing. I'd rather not add keyset-pagination surgery to membership-mutating SQL to defend against a blowup this query shape doesn't actually produce. If real scale ever demands it, the existing range-endpoint pattern is the clean follow-up.

includes/cli/class-integrity-check.php
adekbadek: Addressed in 9cdc901. The progress bar now ticks once per actionable item regardless of whether it dispatched, and items skipped for missing data (no hub lookup / empty `membership_id`) are counted in the skipped total, so the bar always finishes at 100%.

includes/cli/class-integrity-check.php
adekbadek: Addressed in 9cdc901. `parse_gmt_timestamp()` now checks `getLastErrors()` and rejects non-positive results, so WP zero-dates (`0000-00-00 00:00:00`) return `false` and the caller's `time()` fallback engages instead of storing a large-negative timestamp.

includes/node/class-integrity-check-endpoints.php
jason10lee: Good spot on the mixup: the right query looks like a plain JOIN, and what remains shouldn't blow up.  Withdrawn!

jason10lee (approved):
All three blockers addressed and confirmed working on a live env, plus additional small items handled with no further comment from agentic review.  Approved!

PR #319, opened 2026-05-11
fix(network): harden the user-sync receive side and sign the Hub pull response

adekbadek:

Two pieces of hardening on the Hub/Node sync, neither of which changes what the sync features do (automatic profile/role sync and the "Sync user across network" button — including an admin propagating an admin — all work as before):

- **(a) receive-side robustness on the user-sync event handlers**, and
- **(b) man-in-the-middle hardening of the pull channel.**

Note on scope: a Newspack Network treats all member sites as trusted participants, so this PR is *not* about defending against a malicious Node. It's about (a) not letting a bug or plugin interaction on a Node push fields the sync was never meant to carry, and (b) not letting an attacker *on the wire between a Node and the Hub* (a proxy/CDN/WAF in the path, a TLS misconfig, DNS hijack) inject events — a wire attacker isn't a Node.

**1. Allowlist incoming `prop` / `meta`** (`User_Updated`, `User_Manually_Synced`)

`maybe_update_user()` / `maybe_sync_user()` copied every `data->prop` into `wp_update_user()` and every `data->meta` into `update_user_meta()` with no allowlist — so a malformed/bug-produced payload could write things like `user_pass`, `wp_capabilities`, `_application_passwords`, `np_reader_email_verified`, none of which the sync legitimately emits. Now restricted to `User_Update_Watcher::$user_props` and `User_Update_Watcher::get_writable_meta()` — exactly the keys the legitimate sync produces (`User_Update_Watcher` for `network_user_updated`, `User_Manual_Sync::manual_sync_user()` for `network_manual_sync_user`). This is the allowlist already used by `class-author-ingestion.php` and `class-incoming-post.php`; these two handlers were the only user-writing paths that didn't use it. `wp_update_user()` is also skipped when no allowed prop is present (the array would contain only `ID`). Role propagation and auto-create in `User_Manually_Synced` are untouched.

**2. Avatar sideload is SSRF-guarded** (`Users::maybe_sideload_avatar()`)

The avatar URL comes from the event payload and is fetched server-side (`media_sideload_image()` → `download_url()` → `wp_remote_get()`, which does not block private/reserved hosts). It's now validated with `wp_http_validate_url()` before sideloading. Fixed in the shared util, so the author-ingestion and incoming-post callers are covered too. (Used `wp_http_validate_url()` rather than scoping to the originating host so legitimately CDN'd avatar URLs — e.g. Jetpack Site Accelerator rewrites attachment URLs to `i*.wp.com` — keep working.)

**3. The Hub pull response is signed; request params are authenticated** (`Pull_Endpoint`, `Pulling`, `Requests`)

The Node's pull *request* was signed, but `Pull_Endpoint::handle_pull()` returned the event list in cleartext and read `actions` / `last_processed_id` from the *plaintext* request params (not the signed payload). A man-in-the-middle on the Node↔Hub link could therefore rewrite the response to inject any pullable event (e.g. `network_manual_sync_user` creating an administrator) or tamper with the request. Now:

- `Requests::get_request_to_hub_errors()` returns the decrypted request parameters; `handle_pull()` reads `actions` / `last_processed_id` / the signed-response flag from those, not the spoofable plaintext.
- The Node always asks for a signed response; the Hub encrypts the response body with the shared secret; the Node decrypts it and refuses to process an unsigned or unverifiable response. Older Nodes that don't set the flag still get the plain body, so a freshly-updated Hub doesn't break them.

1. Set up a Hub + Node (`n setup` on two sites, one hub, one node). Confirm normal sync still works: register a reader on the Node → propagates to the Hub; edit a reader's profile / social link / billing / email → propagates.
2. Edit an admin on the Node and click **Sync user across network** — confirm the user and their role (`administrator`) propagate and the account is created on peers. Also confirm an admin's profile edits (incl. email) propagate via the automatic `network_user_updated` event.
3. Trigger a Node pull (wait for the cron or use the manual pull button); confirm events still flow. Then point the Node's hub URL at something that returns a cleartext `{data:[...],total:N}` body — confirm the Node logs an "unsigned pull response" error and processes nothing.
4. Send a crafted `network_user_updated` webhook to the Hub for an existing user's email with `prop: { user_pass, … }` and `meta: { _application_passwords, np_reader_email_verified, wp_capabilities, … }` — confirm none of those keys are written; the `User_Update_Watcher` profile/meta keys still are.
5. Run `n test-php` in `newspack-network`.

### Other information:

- PHPCS passes on the changed files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/includes/hub/class-pull-endpoint.php
+++ b/includes/hub/class-pull-endpoint.php
@@ -8,6 +8,7 @@
 namespace Newspack_Network\Hub;
 
 use Newspack_Network\Accepted_Actions;
+use Newspack_Network\Crypto;
 use Newspack_Network\Debugger;
 use Newspack_Network\Hub\Stores\Event_Log;
 use WP_REST_Response;
@@ -72,14 +73,19 @@ class Pull_Endpoint {
 	 * @return WP_REST_Response
 	 */
 	public static function handle_pull( $request ) {
-		$request_error = \Newspack_Network\Utils\Requests::get_request_to_hub_errors( $request );
-		if ( \is_wp_error( $request_error ) ) {
-			return new WP_REST_Response( [ 'error' => $request_error->get_error_message() ], 403 );
+		$verified_params = \Newspack_Network\Utils\Requests::get_request_to_hub_errors( $request );
+		if ( \is_wp_error( $verified_params ) ) {
+			return new WP_REST_Response( [ 'error' => $verified_params->get_error_message() ], 403 );
 		}
 
+		// Read the request parameters from the verified (signed) payload, not the plaintext
+		// copies, so a man-in-the-middle can't tamper with the action list, the cursor, or
+		// the signed-response flag. The plaintext 'site' is fine to use for the Node lookup:
+		// it selected the secret that decrypted the signature, so it identifies the sender.
 		$site              = $request['site'];
-		$last_processed_id = $request['last_processed_id'];
-		$actions           = $request['actions'];
+		$last_processed_id = (int) ( $verified_params['last_processed_id'] ?? 0 );
+		$actions           = (array) ( $verified_params['actions'] ?? [] );
+		$signed_response   = ! empty( $verified_params['signed_response'] );
 
 		Debugger::log( sprintf( 'Pull request received from site %s, with last processed ID %d, for actions: %s.', $site, $last_processed_id, implode( ', ', $actions ) ) );
 
@@ -122,6 +128,25 @@ class Pull_Endpoint {
 			'data'  => $events_formatted,
 			'total' => $total_events,
 		];
+
+		// When the Node asked for a signed response, encrypt the body with the shared secret
+		// so a man-in-the-middle can't inject events into it. Older Nodes don't set the flag
+		// and get the body unencrypted, as before.
+		if ( $signed_response ) {
+			$response_nonce  = Crypto::generate_nonce();
+			$encrypted_body  = Crypto::encrypt_message( wp_json_encode( $response_body ), $node->get_secret_key(), $response_nonce );
+			if ( is_wp_error( $encrypted_body ) || ! is_string( $encrypted_body ) ) {
+				Debugger::log( 'Could not sign the pull response.' );
+				return new WP_REST_Response( [ 'error' => 'Could not sign response.' ], 500 );
+			}
+			return new WP_REST_Response(
+				[
+					'nonce' => $response_nonce,
+					'data'  => $encrypted_body,
+				]
+			);
+		}
+
 		return new WP_REST_Response( $response_body );
 	}
 }
--- a/includes/hub/class-pull-endpoint.php
+++ b/includes/hub/class-pull-endpoint.php
@@ -87,6 +87,12 @@ class Pull_Endpoint {
 		$actions           = (array) ( $verified_params['actions'] ?? [] );
 		$signed_response   = ! empty( $verified_params['signed_response'] );
 
+		// Defense-in-depth: cross-check the plaintext 'site' against the signed copy so a
+		// mismatch is rejected even if a future change makes the lookup tolerant of either.
+		if ( isset( $verified_params['site'] ) && $verified_params['site'] !== $site ) {
+			return new WP_REST_Response( [ 'error' => 'Site mismatch.' ], 403 );
+		}
+
 		Debugger::log( sprintf( 'Pull request received from site %s, with last processed ID %d, for actions: %s.', $site, $last_processed_id, implode( ', ', $actions ) ) );
 
 		if ( empty( $actions ) ) {
--- a/includes/hub/class-pull-endpoint.php
+++ b/includes/hub/class-pull-endpoint.php
@@ -73,7 +73,7 @@ class Pull_Endpoint {
 	 * @return WP_REST_Response
 	 */
 	public static function handle_pull( $request ) {
-		$verified_params = \Newspack_Network\Utils\Requests::get_request_to_hub_errors( $request );
+		$verified_params = \Newspack_Network\Utils\Requests::verify_request_to_hub( $request );
 		if ( \is_wp_error( $verified_params ) ) {
 			return new WP_REST_Response( [ 'error' => $verified_params->get_error_message() ], 403 );
 		}
@@ -133,8 +133,13 @@ class Pull_Endpoint {
 		// so a man-in-the-middle can't inject events into it. Older Nodes don't set the flag
 		// and get the body unencrypted, as before.
 		if ( $signed_response ) {
-			$response_nonce  = Crypto::generate_nonce();
-			$encrypted_body  = Crypto::encrypt_message( wp_json_encode( $response_body ), $node->get_secret_key(), $response_nonce );
+			$response_json = wp_json_encode( $response_body );
+			if ( false === $response_json ) {
+				Debugger::log( 'Could not encode the pull response.' );
+				return new WP_REST_Response( [ 'error' => 'Could not encode response.' ], 500 );
+			}
+			$response_nonce = Crypto::generate_nonce();
+			$encrypted_body = Crypto::encrypt_message( $response_json, $node->get_secret_key(), $response_nonce );
 			if ( is_wp_error( $encrypted_body ) || ! is_string( $encrypted_body ) ) {
 				Debugger::log( 'Could not sign the pull response.' );
 				return new WP_REST_Response( [ 'error' => 'Could not sign response.' ], 500 );
--- a/includes/incoming-events/class-user-manually-synced.php
+++ b/includes/incoming-events/class-user-manually-synced.php
@@ -44,7 +44,7 @@ class User_Manually_Synced extends Abstract_Incoming_Event {
 	 */
 	public function maybe_sync_user() {
 		$email = $this->get_email();
-		Debugger::log( 'Processing user_manually_synced with email: ' . $email );
+		Debugger::log( sprintf( 'Processing user_manually_synced for %s from %s.', $email, $this->get_site() ) );
 		if ( ! $email ) {
 			return;
 		}
--- a/includes/incoming-events/class-user-manually-synced.php
+++ b/includes/incoming-events/class-user-manually-synced.php
@@ -104,13 +104,19 @@ class User_Manually_Synced extends Abstract_Incoming_Event {
 			}
 		}
 
-		// Loop through user props and update.
+		// Loop through user props and update. Roles are synced above via $data->role; the
+		// props and meta below are restricted to the fixed set of fields the user sync
+		// tracks, so an incoming payload can't write arbitrary props (user_pass, ...) or
+		// meta (_application_passwords, wp_capabilities, ...).
 		if ( isset( $data->prop ) ) {
-			$update_array = [
+			$incoming_props = (array) $data->prop;
+			$update_array   = [
 				'ID' => $user->ID,
 			];
-			foreach ( $data->prop as $prop_key => $prop_value ) {
-				$update_array[ $prop_key ] = $prop_value;
+			foreach ( User_Update_Watcher::$user_props as $prop_key ) {
+				if ( isset( $incoming_props[ $prop_key ] ) ) {
+					$update_array[ $prop_key ] = $incoming_props[ $prop_key ];
+				}
 			}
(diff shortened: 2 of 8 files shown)


includes/incoming-events/class-user-updated.php
```diff
@@ -54,26 +54,44 @@ public function maybe_update_user() {
...
+		// _application_passwords) must never reach wp_update_user() / update_user_meta().
 		if ( isset( $data->prop ) ) {
-			$update_array = [
+			$incoming_props = (array) $data->prop;
+			$update_array   = [
 				'ID' => $existing_user->ID,
 			];
-			foreach ( $data->prop as $prop_key => $prop_value ) {
-				$update_array[ $prop_key ] = $prop_value;
+			foreach ( User_Update_Watcher::$user_props as $prop_key ) {
```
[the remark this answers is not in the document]
adekbadek: Addressed in 384b756 — `wp_update_user()` is now only called when at least one allowed prop is present (the array would otherwise contain just `ID`). Same in `User_Manually_Synced`.

includes/incoming-events/class-user-updated.php
```diff
@@ -54,26 +54,44 @@ public function maybe_update_user() {
...
 
 			wp_update_user( $update_array );
 		}
 
 		if ( isset( $data->meta ) ) {
-			foreach ( $data->meta as $meta_key => $meta_value ) {
-				Debugger::log( 'Updating user meta: ' . $meta_key );
-				update_user_meta( $existing_user->ID, $meta_key, $meta_value );
+			$incoming_meta = (array) $data->meta;
+			foreach ( User_Update_Watcher::get_writable_meta() as $meta_key ) {
```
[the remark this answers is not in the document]
adekbadek: Addressed in 384b756. Went with `wp_http_validate_url()` on the avatar URL before sideloading rather than scoping to the originating host: `media_sideload_image()` -> `download_url()` -> `wp_remote_get()` doesn't block private/reserved hosts, so that's the SSRF concern, and host-scoping would break legitimately CDN'd avatar URLs (e.g. Jetpack Site Accelerator rewrites attachment URLs to `i*.wp.com`). Fixed in `Users::maybe_sideload_avatar()` so the author-ingestion and incoming-post callers are covered too.

includes/utils/class-requests.php
```diff
@@ -76,13 +80,13 @@ public static function get_request_to_hub_errors( $request ) {
...
 		}
 
-		$verified         = $node->decrypt_message( $signature, $nonce );
-		$verified_message = json_decode( $verified );
-		if ( ! $verified || ! is_object( $verified_message ) ) {
+		$verified = $node->decrypt_message( $signature, $nonce );
+		$params   = is_string( $verified ) ? json_decode( $verified, true ) : null;
+		if ( ! is_array( $params ) ) {
 			\Newspack_Network\Debugger::log( 'Signature check failed' );
 			return new WP_Error( 'newspack_network_bad_request_signature', __( 'Bad request. Invalid signature.', 'newspack-network' ) );
```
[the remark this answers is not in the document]
adekbadek: Addressed in 87c4bb5 — hardened `Crypto::decrypt_message()` instead: it now rejects non-hex / odd-length inputs up front and catches `\Throwable` (so a `TypeError` from `hex2bin()` returning `false` no longer escapes), returning `false`. `verify_request_to_hub()` then surfaces that as a 403, not a 500. Covers the `/webhook` handler too.

includes/node/class-pulling.php
```diff
@@ -226,7 +227,24 @@ public static function pull() {
...
+		// response, and must not be processed.
+		if ( ! is_array( $response ) || empty( $response['nonce'] ) || ! isset( $response['data'] ) || ! is_string( $response['data'] ) ) {
+			self::handle_error( new \WP_Error( 'newspack-network-node-pull-unsigned-response', __( 'The Hub returned an unsigned pull response. Make sure the Hub is up to date.', 'newspack-network' ) ) );
+			return;
+		}
+
+		$verified = Crypto::decrypt_message( $response['data'], Settings::get_secret_key(), $response['nonce'] );
+		if ( false === $verified ) {
+			self::handle_error( new \WP_Error( 'newspack-network-node-pull-signature', __( 'Could not verify the Hub pull response signature.', 'newspack-network' ) ) );
+			return;
```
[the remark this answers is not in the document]
adekbadek: Addressed in 87c4bb5 via the same `Crypto::decrypt_message()` hardening — a tampered response with non-hex `nonce`/`data` now makes `decrypt_message()` return `false`, which `pull()` already handles by logging an unverifiable-response error and bailing. No fatal.

includes/hub/class-pull-endpoint.php
```diff
@@ -122,6 +128,25 @@ function( $event ) {
 			'data'  => $events_formatted,
 			'total' => $total_events,
 		];
+
+		// When the Node asked for a signed response, encrypt the body with the shared secret
+		// so a man-in-the-middle can't inject events into it. Older Nodes don't set the flag
+		// and get the body unencrypted, as before.
+		if ( $signed_response ) {
+			$response_nonce  = Crypto::generate_nonce();
+			$encrypted_body  = Crypto::encrypt_message( wp_json_encode( $response_body ), $node->get_secret_key(), $response_nonce );
```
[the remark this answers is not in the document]
adekbadek: Addressed in 87c4bb5 — `handle_pull()` now checks `wp_json_encode()` didn't return `false` and 500s cleanly if it did, and `Crypto::encrypt_message()` also rejects a non-string message and catches `\Throwable` as a backstop.

includes/utils/class-users.php
```diff
@@ -132,9 +132,19 @@ public static function maybe_sideload_avatar( $user_id, $user_data, $overwrite )
 		}
 
 		if ( array_key_exists( $avatar_meta_key, $user_data ) && ! empty( $user_data[ $avatar_meta_key ]['full'] ) ) {
-			Debugger::log( 'Updating user avatar' );
 			$avatar_url = $user_data[ $avatar_meta_key ]['full'];
```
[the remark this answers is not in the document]
adekbadek: Addressed in 87c4bb5 — the guard now requires `is_array( $user_data[ $avatar_meta_key ] )` before reading `['full']`, then `wp_http_validate_url()` on the resulting string (the `is_string()` check on the URL was already there).

includes/utils/class-requests.php
```diff
@@ -52,10 +52,14 @@ public static function sign_params( $params ) {
...
+	 * Validate a request to the Hub and return its authenticated parameters.
+	 *
+	 * The signature is verified against the requesting Node's secret key; on success the
+	 * decrypted request parameters are returned so callers can use those instead of the
+	 * plaintext copies (which a man-in-the-middle could tamper with).
 	 *
 	 * @param WP_REST_Request $request Full data about the request.
-	 * @return bool|WP_Error True if the request is valid, WP_Error otherwise.
+	 * @return array|WP_Error The verified request parameters, or WP_Error if verification failed.
 	 */
```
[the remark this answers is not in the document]
adekbadek: Addressed in 87c4bb5 — renamed to `verify_request_to_hub()`. One internal caller, updated; no deprecated wrapper needed.

includes/node/class-pulling.php
dkoo: ### 🟡 Suggestion: Pull response is authenticated but not replay-protected

`process_pulled_data()` calls `set_last_processed_id($id)` unconditionally. AEAD prevents an attacker from forging a response, but a previously-captured ciphertext+nonce can be replayed: events get reprocessed and the cursor can roll **backward** (next pull then re-fetches everything from there). 

Replay is strictly worse before this PR (no auth at all on the response), so this isn't a regression — but it would be in-scope for the PR's security hardening. Cheapest mitigation: clamp `set_last_processed_id($id)` to `max(current, $id)` and skip events with `id ≤ current`. Stronger: bind a Node-generated request nonce into the signed response and reject mismatches.

includes/hub/class-pull-endpoint.php
```diff
@@ -72,14 +73,19 @@ public static function get_pull_limit() {
...
+		$verified_params = \Newspack_Network\Utils\Requests::verify_request_to_hub( $request );
+		if ( \is_wp_error( $verified_params ) ) {
+			return new WP_REST_Response( [ 'error' => $verified_params->get_error_message() ], 403 );
 		}
 
+		// Read the request parameters from the verified (signed) payload, not the plaintext
+		// copies, so a man-in-the-middle can't tamper with the action list, the cursor, or
+		// the signed-response flag. The plaintext 'site' is fine to use for the Node lookup:
+		// it selected the secret that decrypted the signature, so it identifies the sender.
 		$site              = $request['site'];
```
dkoo, line 85: ### 🟡 Suggestion: Defense-in-depth: cross-check `site` between plaintext and signed payload

The plaintext `site` is used to look up the node's key, which is fine (wrong key = decryption fails), but an additional check `$verified_params['site'] === $request['site']` prevents future drift.

includes/node/class-pulling.php
```diff
@@ -226,7 +227,24 @@ public static function pull() {
...
+		}
+
+		$response = json_decode( $verified, true );
```
dkoo, line 263: ### 🟡 Suggestion: Silent fallthrough on a signed-but-malformed body

If `json_decode($verified)` returns `null` or the structure is wrong, the code silently returns. Add a `handle_error()` for symmetry with the unsigned/unverifiable branches.

includes/incoming-events/class-user-manually-synced.php
```diff
@@ -44,7 +44,7 @@ public function process_in_node() {
 	 */
 	public function maybe_sync_user() {
```
dkoo, line 53: ### 🟡 Suggestion: Specify why role and `user_login` paths in `User_Manually_Synced` are still unconstrained

`$data->role` (set on [L83](https://github.com/Automattic/newspack-network/pull/319/changes#diff-ccce892f907c03a4a37ce5584a1875023255e21d08dae968c586e5da0b1a373bR83)) can grant `administrator`; `$data->user_login` (set on [L59](https://github.com/Automattic/newspack-network/pull/319/changes#diff-ccce892f907c03a4a37ce5584a1875023255e21d08dae968c586e5da0b1a373bR59)) reaches `get_or_create_user_by_email()` without `sanitize_user()`. Pre-existing, in-scope given the PR's framing — but the PR description explicitly notes role propagation is intended ("admin propagating an admin"), so this is a deliberate trust-the-Node decision under the stated threat model. Worth a comment but not a fix.

dkoo (approved):
All manual testing steps pass as described, and the additional security measures are a welcome improvement over `trunk`. No bugs or regressions introduced by the changes, so approving with some inline suggestions below.

Also consider my prior question about whether this warrants a hotfix instead of being rolled into a future feature release.

includes/node/class-pulling.php
adekbadek: Addressed in 7f75064 — `process_pulled_data` now reads the current cursor up front, skips events with `id <= current`, tracks the highest ID processed in the batch, and only writes the option if it would advance. A replayed envelope can no longer roll the cursor backward or reprocess events.

includes/hub/class-pull-endpoint.php
adekbadek: Addressed in 7f75064 — the Hub pull endpoint now rejects a request whose signed `site` doesn't match the plaintext `site` (403). Decryption keying off the plaintext `site` already catches a substituted key, but the explicit check guards against future drift.

includes/node/class-pulling.php
adekbadek: Addressed in 7f75064 — a signed-but-malformed pull body (non-array decode or missing `data` array) now goes through `handle_error()` with a `newspack-network-node-pull-malformed-response` code, matching the unsigned/unverifiable branches.

includes/incoming-events/class-user-manually-synced.php
adekbadek: Addressed in 7f75064 — added a docblock on `maybe_sync_user()` calling out that `$data->role` (including `administrator`) and `$data->user_login` are intentionally accepted unconstrained under the current "all Nodes are operated by the same operator" trust model, with a note about what would need to change (sanitize `user_login` via `sanitize_user()`, restrict `role` to a safe set) if that ever shifts.

